PR #46, opened 2023-03-20
OSD-15567: Add Proxy URL support to login  command

samanthajayasinghe:
### What type of PR is this?

feature

Enable proxy URL support to  `login` command

https://issues.redhat.com/browse/OSD-15567

- [x] Ran unit tests locally

### DIFF
--- a/cmd/ocm-backplane/cloud/console.go
+++ b/cmd/ocm-backplane/cloud/console.go
@@ -14,6 +14,7 @@ import (
 	"github.com/spf13/cobra"
 
 	BackplaneApi "github.com/openshift/backplane-api/pkg/client"
+	"github.com/openshift/backplane-cli/pkg/cli/config"
 	"github.com/openshift/backplane-cli/pkg/utils"
 )
 
@@ -110,7 +111,9 @@ func runConsole(cmd *cobra.Command, argv []string) (err error) {
 
 	// ============Get Backplane URl ==========================
 	if consoleArgs.backplaneURL == "" {
-		consoleArgs.backplaneURL, err = utils.DefaultOCMInterface.GetBackplaneURL()
+		// Get Backplane configuration
+		bpConfig, err := config.GetBackplaneConfiguration()
+		consoleArgs.backplaneURL = bpConfig.URL
 		if err != nil || consoleArgs.backplaneURL == "" {
 			return fmt.Errorf("can't find backplane url: %w", err)
 		}
--- a/cmd/ocm-backplane/cloud/console.go
+++ b/cmd/ocm-backplane/cloud/console.go
@@ -110,18 +110,21 @@ func runConsole(cmd *cobra.Command, argv []string) (err error) {
 		"Name": clusterName}).Infoln("Target cluster")
 
 	// ============Get Backplane URl ==========================
-	if consoleArgs.backplaneURL == "" {
+	bpURL := ""
+	if consoleArgs.backplaneURL != "" {
+		bpURL = consoleArgs.backplaneURL
+	} else {
 		// Get Backplane configuration
 		bpConfig, err := config.GetBackplaneConfiguration()
-		consoleArgs.backplaneURL = bpConfig.URL
-		if err != nil || consoleArgs.backplaneURL == "" {
+		if err != nil || bpConfig.URL == "" {
 			return fmt.Errorf("can't find backplane url: %w", err)
 		}
-		logger.Infof("Using backplane URL: %s\n", consoleArgs.backplaneURL)
+		bpURL = bpConfig.URL
+		logger.Infof("Using backplane URL: %s\n", bpURL)
 	}
 
 	// ======== Get cloudconsole from backplane API ============
-	response, err := getCloudConsole(consoleArgs.backplaneURL, clusterId)
+	response, err := getCloudConsole(bpURL, clusterId)
 	if err != nil {
 		return err
 	}
--- a/cmd/ocm-backplane/cloud/credentials.go
+++ b/cmd/ocm-backplane/cloud/credentials.go
@@ -141,20 +141,23 @@ func runCredentials(cmd *cobra.Command, argv []string) error {
 		"Name": clusterName}).Infoln("Target cluster")
 
 	// ============Get Backplane URl ==========================
-	if credentialArgs.backplaneURL == "" {
+	bpURL := ""
+	if consoleArgs.backplaneURL != "" {
+		bpURL = credentialArgs.backplaneURL
+	} else {
 		// Get Backplane configuration
 		bpConfig, err := config.GetBackplaneConfiguration()
-		credentialArgs.backplaneURL = bpConfig.URL
-		if err != nil || credentialArgs.backplaneURL == "" {
+		if err != nil || bpConfig.URL == "" {
 			return fmt.Errorf("can't find backplane url: %w", err)
 		}
-		logger.Infof("Using backplane URL: %s\n", credentialArgs.backplaneURL)
+		bpURL = bpConfig.URL
+		logger.Infof("Using backplane URL: %s\n", bpURL)
 	}
 
 	// ======== Call Endpoint ==================================
 	logger.Debugln("Getting Cloud Credentials")
 
-	credsResp, _ := getCloudCredential(credentialArgs.backplaneURL, clusterId)
+	credsResp, _ := getCloudCredential(bpURL, clusterId)
 
 	// ======== Render cloud credentials =======================
 	switch cloudProvider {
--- a/cmd/ocm-backplane/cloud/credentials.go
+++ b/cmd/ocm-backplane/cloud/credentials.go
@@ -11,6 +11,7 @@ import (
 	"github.com/spf13/cobra"
 
 	BackplaneApi "github.com/openshift/backplane-api/pkg/client"
+	"github.com/openshift/backplane-cli/pkg/cli/config"
 	"github.com/openshift/backplane-cli/pkg/utils"
 )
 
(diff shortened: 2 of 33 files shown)


cmd/ocm-backplane/login/login.go
```diff
@@ -79,6 +87,13 @@ func runLogin(cmd *cobra.Command, argv []string) (err error) {
 		clusterKey = clusterInfo.ClusterID
 	}
 
+	// Init Proxy url
+	err = utils.DefaultOCMInterface.SetBackplaneProxyURL(args.proxyURL)
```
wanghaoran1988, line 91: We should only set if proxyURL is not empty

cmd/ocm-backplane/login/login.go
```diff
@@ -152,6 +167,12 @@ func runLogin(cmd *cobra.Command, argv []string) (err error) {
 
 	targetCluster.Server = proxyURL
 
+	// Add proxy URL to target cluster
+	proxyURL, hasProxyUrl := os.LookupEnv(info.BACKPLANE_PROXY_ENV_NAME)
+	if hasProxyUrl {
+		targetCluster.ProxyURL = proxyURL
+	}
+
```
wanghaoran1988, line 175: what if the proxy is set in the config file, but not specified using --proxy ?
it looks like we should do like this:
1. If --proxy is used, use this proxy url and write it to kubeconfig
2. We should ignore that? since after login, you're already using this proxy when you run oc command
3. If the config already set this, read this proxy url and write to kubeconfig when login.

cmd/ocm-backplane/login/login.go
samanthajayasinghe: yes . always check proxy URL is not empty before setting it to the env vars 
 https://github.com/openshift/backplane-cli/blob/67e50c71be30aad3c328a63cb3972ea2438bb9fc/pkg/utils/ocmWrapper.go#L270

pkg/utils/ocmWrapper.go
```diff
@@ -222,6 +224,55 @@ func (*DefaultOCMInterfaceImpl) GetBackplaneURL() (string, error) {
 	return "", nil
 }
 
+// Set Backplane Proxy URL from local settings
+func (*DefaultOCMInterfaceImpl) SetBackplaneProxyURL(proxyUrl string) error {
```
wanghaoran1988, line 228: We can remove this

cmd/ocm-backplane/login/login.go
samanthajayasinghe: If --proxy is not used, then proxy derived based on the following order 
1. via env vars 
https://github.com/openshift/backplane-cli/blob/67e50c71be30aad3c328a63cb3972ea2438bb9fc/pkg/utils/ocmWrapper.go#L237
2. `backplane.json` file 
https://github.com/openshift/backplane-cli/blob/67e50c71be30aad3c328a63cb3972ea2438bb9fc/pkg/utils/ocmWrapper.go#L245

I always write `proxy` into .kubeconfig if it's not empty. 

Am I missing something?

pkg/utils/ocmWrapper.go
samanthajayasinghe: Ack .. Shall I move this proxy URL deriving logic to the utility function then?

samanthajayasinghe: done

cmd/ocm-backplane/login/login.go
samanthajayasinghe: Done

cmd/ocm-backplane/login/login.go
samanthajayasinghe: As we discussed, This case is handled vis `http.transport`

pkg/cli/config/config.go
```diff
@@ -0,0 +1,58 @@
...
+	if _, err := os.Stat(filePath); err == nil {
+		file, err := os.Open(filePath)
+
+		if err != nil {
+			return proxyUrl, fmt.Errorf("failed to read file %s : %v", filePath, err)
+		}
+
+		defer file.Close()
+		decoder := json.NewDecoder(file)
+		configuration := BackplaneConfiguration{}
```
supreeth7, line 41: Rename `configuration` -> `bpConfig` ?

cmd/ocm-backplane/login/login.go
```diff
@@ -15,41 +15,42 @@ import (
...
-	Long: `Running login command will send a request to backplane api
-	using OCM token. The backplane api will return a proxy url for 
-	target cluster. The url will be written to kubeconfig, so we can 
-	run oc command later to operate the target cluster.`,
-	Example:      " backplane login <id>\n backplane login %test%\n backplane login <external_id>",
-	Args:         cobra.ExactArgs(1),
-	RunE:         runLogin,
-	SilenceUsage: true,
-}
+	globalOpts = &globalflags.GlobalOptions{}
```
supreeth7, line 31: We can perhaps leverage Cobra's `rootCmd.PersistentFlags()` to set the global options in a better way.

Ref: https://github.com/spf13/cobra/blob/main/user_guide.md#persistent-flags

pkg/cli/config/config.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/openshift/backplane-cli/pkg/info"
+)
+
+type BackplaneConfiguration struct {
+	URL       string
+	Proxy_URL string
+}
+
+func GetBackplaneConfigFile() string {
+	userHomeDir, _ := os.UserHomeDir()
```
supreeth7, line 18: Instead of storing the configuration in user's `$HOME`, `$HOME/.config` would be an ideal place to store config files I suppose?

```suggestion
	userConfigDir, _ := os.UserConfigDir()
```

cmd/ocm-backplane/login/login.go
samanthajayasinghe: Done

pkg/cli/config/config.go
samanthajayasinghe: Done

cmd/ocm-backplane/login/login.go
```diff
@@ -101,12 +117,12 @@ func runLogin(cmd *cobra.Command, argv []string) (err error) {
 	}
 
 	// Get Backplane URL
-	if args.backplaneURL == "" {
-		args.backplaneURL, err = utils.DefaultOCMInterface.GetBackplaneURL()
-		if err != nil || args.backplaneURL == "" {
+	if globalOpts.BackplaneURL == "" {
+		globalOpts.BackplaneURL, err = utils.DefaultOCMInterface.GetBackplaneURL()
```
wanghaoran1988, line 121: We should move this `GetBackplaneURL` func out of OCM interface? it's not part of OCM.

pkg/cli/config/config.go
```diff
@@ -0,0 +1,67 @@
...
+
+	if err != nil {
+		return fileName
+	}
+
+	return configDir + "/" + fileName
+}
+
+// Get Backplane ProxyUrl from config
+func GetBackplaneProxyUrl() (proxyUrl string, err error) {
```
wanghaoran1988, line 37: Rename this to GetBackplaneConfiguration  and return the config struct dirrectly?

pkg/utils/clientUtils.go
```diff
@@ -16,13 +18,17 @@ type ClientUtils interface {
 	MakeRawBackplaneAPIClientWithAccessToken(base, accessToken string) (BackplaneApi.ClientInterface, error)
 	MakeRawBackplaneAPIClient(base string) (BackplaneApi.ClientInterface, error)
 	GetBackplaneClient(backplaneURL string) (client BackplaneApi.ClientInterface, err error)
+	SetClientProxyUrl(proxyUrl string) error
 }
 
 type DefaultClientUtilsImpl struct{}
```
wanghaoran1988, line 24:
```suggestion
type DefaultClientUtilsImpl struct{
    clientProxyUrl string
}
```

cmd/ocm-backplane/login/login.go
samanthajayasinghe: Done

cmd/ocm-backplane/cloud/console.go
```diff
@@ -110,7 +111,9 @@ func runConsole(cmd *cobra.Command, argv []string) (err error) {
 
 	// ============Get Backplane URl ==========================
 	if consoleArgs.backplaneURL == "" {
-		consoleArgs.backplaneURL, err = utils.DefaultOCMInterface.GetBackplaneURL()
+		// Get Backplane configuration
+		bpConfig, err := config.GetBackplaneConfiguration()
+		consoleArgs.backplaneURL = bpConfig.URL
 		if err != nil || consoleArgs.backplaneURL == "" {
 			return fmt.Errorf("can't find backplane url: %w", err)
 		}
```
wanghaoran1988, line 119:
```suggestion
		}
```

cmd/ocm-backplane/cloud/console.go
```diff
@@ -110,7 +111,9 @@ func runConsole(cmd *cobra.Command, argv []string) (err error) {
 
 	// ============Get Backplane URl ==========================
 	if consoleArgs.backplaneURL == "" {
```
wanghaoran1988, line 113: How about this:
```
	bpURL := ""
	if consoleArgs.backplaneURL != "" {
		bpURL = consoleArgs.backplaneURL
	} else {
		// Get Backplane configuration
		bpConfig, err := config.GetBackplaneConfiguration()
		if err != nil || bpConfig.URL == "" {
			return fmt.Errorf("can't find backplane url: %w", err)
		}
		bpURL = bpConfig.URL
		logger.Infof("Using backplane URL: %s\n", bpURL)
	}

	// ======== Get cloudconsole from backplane API ============
	response, err := getCloudConsole(bpURL, clusterId)
	if err != nil {
		return err
	}
```

cmd/ocm-backplane/cloud/credentials.go
```diff
@@ -141,7 +142,9 @@ func runCredentials(cmd *cobra.Command, argv []string) error {
 
 	// ============Get Backplane URl ==========================
 	if credentialArgs.backplaneURL == "" {
-		credentialArgs.backplaneURL, err = utils.DefaultOCMInterface.GetBackplaneURL()
+		// Get Backplane configuration
+		bpConfig, err := config.GetBackplaneConfiguration()
```
wanghaoran1988, line 146: Ditto

cmd/ocm-backplane/cloud/console.go
samanthajayasinghe: done

cmd/ocm-backplane/cloud/credentials.go
samanthajayasinghe: done

PR #53, opened 2023-03-27
OSD-15495 - Add method to check if the container is already running

Tafhim:
### What type of PR is this?

bug

We need this PR to fix console container creation issues when console is already running.

[OSD-15495](https://issues.redhat.com//browse/OSD-15495)

### DIFF
--- a/cmd/ocm-backplane/console/console.go
+++ b/cmd/ocm-backplane/console/console.go
@@ -130,12 +130,12 @@ func init() {
 	)
 }
 
-func checkAndFindContainerURL(containerName string, containerEngine string) (err error) {
+func checkContainerExists(containerName string, containerEngine string) (exists bool, err error) {
 	existCheckArgs := []string{
 		"container",
 		"ps",
 		"--filter",
-		fmt.Sprintf("\"name=%s\"", containerName),
+		fmt.Sprintf("name=%s", containerName),
 	}
 
 	existCheckCmd, existCheckCmdOutput := createCommand(containerEngine, existCheckArgs...), new(strings.Builder)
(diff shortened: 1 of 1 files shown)


cmd/ocm-backplane/console/console.go
```diff
@@ -129,6 +130,57 @@ func init() {
...
+		addressCheckCmd.Stdout = addressCheckOutput
+
+		err := addressCheckCmd.Run()
+		if err != nil {
+			return err
+		}
+
+		configurationCmdFragments := []string{}
+
+		json.Unmarshal([]byte(strings.Trim(addressCheckOutput.String(), "\"\n ")), &configurationCmdFragments)
```
supreeth7, line 169: Can you check for any errors here?

cmd/ocm-backplane/console/console.go
```diff
@@ -129,6 +130,57 @@ func init() {
...
+		addressIndex := -1
+		for idx, val := range configurationCmdFragments {
+			if val == "-base-address" {
+				addressIndex = idx + 1
+				break
+			}
+		}
+		return fmt.Errorf("console container is already runninn on %s", configurationCmdFragments[addressIndex])
```
supreeth7, line 177:
```suggestion
		return fmt.Errorf("console container is already running on %s", configurationCmdFragments[addressIndex])
```

supreeth7 (changes requested):

cmd/ocm-backplane/console/console.go
```diff
@@ -129,6 +130,65 @@ func init() {
 	)
 }
 
+func checkAndFindContainerURL(containerName string, containerEngine string) (err error) {
+	existCheckArgs := []string{
+		"container",
+		"ps",
+		"--filter",
```
supreeth7, line 137: The `--filter` here is specific to podman I suppose? For someone using a mac, we use docker and it errors out with:
```
➜ ./ocm-backplane console -b                                                                                    
INFO[0000] Using container engine docker                
Error response from daemon: Invalid filter '"name'
```
Can you try splitting it in two checks based on the container engine?

supreeth7 (changes requested):

PR #56, opened 2023-03-28
Improve test coverage for cloud credentials command

hectorakemp:
### What type of PR is this?

_(cleanup)_

An example implementation of improving a file to the recommended coverage level across the codebase of 80%.  

_Resolves_
https://issues.redhat.com/browse/OSD-15441

- [x] Ran unit tests locally


cmd/ocm-backplane/cloud/credentials.go
```diff
@@ -15,6 +16,13 @@ import (
 	"github.com/openshift/backplane-cli/pkg/utils"
 )
 
+var GetCloudCredential = getCloudCredential
+var GetCloudProvider = getCloudProvider
+var GetBackplaneClusterFromConfig = utils.GetBackplaneClusterFromConfig
+var Unmarshal = json.Unmarshal
+var PrintLn = fmt.Println
+var GetBackplaneConfiguration = config.GetBackplaneConfiguration
```
samanthajayasinghe, line 24: Is it not an anti-pattern of assigning the variable to a function /closure? And it may lose readability too. 
If it needs, then it should be in a common module where everyone can reuse the same vars.

hectorakemp: That's a good point @samanthajayasinghe - do you know any other good ways to isolate methods so that they can be mocked/tested in unit tests? This felt like the cleanest option out of a bunch of quite ugly ones.

hectorakemp: cc @supreeth7 @Tafhim

samanthajayasinghe: My thinking is to add a bunch of standard test methods for `config.GetBackplaneConfiguration` like function and setup via beforeEach()  and clean via afterEach(). 

`utils.GetBackplaneClusterFromConfig` also can be mocked via the temp kubeconfig file and return a fake cluster. 

`getCloudCredential` / `getCloudProvider` already mocked via `ClientUtilsMock` and possible to return fakeResponse or error.

supreeth7: I agree, function assignment isn't a good approach while they might make the code look more structured.

@hectorakemp this overall looks great, can you try mocking `GetBackplaneClusterFromConfig` and make the necessary function calls (`GetBackplaneConfiguration`) before each test case?

supreeth7 (changes requested):
I agree, function assignment isn't a good approach while they might make the code look more structured.

@hectorakemp this overall looks great, can you try mocking `GetBackplaneClusterFromConfig` and make the necessary function calls (`GetBackplaneConfiguration`) before each test case?

pkg/utils/cluster.go
```diff
@@ -16,8 +17,22 @@ type BackplaneCluster struct {
...
+	GetCloudProvider(cluster *cmv1.Cluster) string
+	GetBackplaneCluster(params ...string) (BackplaneCluster, error)
+}
+
+type DefaultClusterUtilsImpl struct {}
+
+var (
+	DefaultClusterUtils ClusterUtils = &DefaultClusterUtilsImpl{}
+)
+
```
samanthajayasinghe, line 33: This is really good and it allows mocking more unit tests :+1:  :1st_place_medal:

hectorakemp: Thanks @samanthajayasinghe ! Ready for another review now I think

cmd/ocm-backplane/cloud/credentials.go
```diff
@@ -15,6 +15,10 @@ import (
 	"github.com/openshift/backplane-cli/pkg/utils"
 )
 
+var GetBackplaneClusterFromConfig = utils.DefaultClusterUtils.GetBackplaneClusterFromConfig
```
samanthajayasinghe, line 18: I guess this 1-1 map completely gets rid of currently since tests are not dependent on these variables.

supreeth7: We can clean this up in another PR since Hector is OOO

supreeth7 (approved):

PR #57, opened 2023-03-28
Change config file name

supreeth7:
### What type of PR is this?
Bug

- Adhere to configuration file naming on UNIX systems.
- Refrain the use of `os.UserConfigDir()` to fetch the configuration file path as in Darwin (MacOS), it returns `$HOME/Library/Application Support` instead of `$HOME/.config`.

- [x] Ran unit tests locally

### DIFF
--- a/pkg/cli/config/config.go
+++ b/pkg/cli/config/config.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/openshift/backplane-cli/pkg/info"
 )
@@ -24,13 +25,14 @@ func GetBackplaneConfigFile() string {
 
 // Get Backplane config default path
 func getConfiDefaultPath(fileName string) string {
-	configDir, err := os.UserConfigDir()
-
+	UserHomeDir, err := os.UserHomeDir()
 	if err != nil {
 		return fileName
 	}
 
-	return configDir + "/" + fileName
+	configFilePath := filepath.Join(UserHomeDir, ".config", fileName)
+
+	return configFilePath
 }
 
 // Get Backplane ProxyUrl from config
--- a/pkg/cli/config/config.go
+++ b/pkg/cli/config/config.go
@@ -14,32 +14,41 @@ type BackplaneConfiguration struct {
 	ProxyURL string `json:"proxy-url"`
 }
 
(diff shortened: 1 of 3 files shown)


iamkirkbater:
+100

TY :)

pkg/cli/config/config.go
```diff
@@ -24,13 +25,14 @@ func GetBackplaneConfigFile() string {
...
 func getConfiDefaultPath(fileName string) string {
-	configDir, err := os.UserConfigDir()
-
+	UserHomeDir, err := os.UserHomeDir()
 	if err != nil {
 		return fileName
 	}
 
-	return configDir + "/" + fileName
+	configFilePath := filepath.Join(UserHomeDir, ".config", fileName)
```
iamkirkbater, line 33: Any chance we could point this to a folder out the gate?  Especially if we are going to be targeting multiple different environments?

So it would ideally be something like:

`$HOME/.config/backplane/config.json`
`$HOME/.config/backplane/config.stage.json`
`$HOME/.config/backplane/config.int.json`

While the stage and int json files could be excluded and would just be personal overrides, having them in a directory gives them all one place to live together, along with any other customizations that the end-user wants to make.

supreeth7: Thanks for the suggestion, this really looks like an ideal way of managing multiple configs, added the changes

pkg/cli/config/config.go
```diff
@@ -24,13 +25,14 @@ func GetBackplaneConfigFile() string {
...
 func getConfiDefaultPath(fileName string) string {
-	configDir, err := os.UserConfigDir()
-
+	UserHomeDir, err := os.UserHomeDir()
 	if err != nil {
 		return fileName
 	}
 
-	return configDir + "/" + fileName
+	configFilePath := filepath.Join(UserHomeDir, info.BACKPLANE_CONFIG_DEFAULT_FILE_PATH, fileName)
```
iamkirkbater, line 33: Since we're pulling the `info` package in here to get the data directly out, we might as well do that for both the default filename and the default config path.  Thoughts?
```suggestion
	configFilePath := filepath.Join(UserHomeDir, info.BACKPLANE_CONFIG_DEFAULT_FILE_PATH, info.BACKPLANE_CONFIG_DEFAULT_FILE_NAME)
```

PR #68, opened 2023-04-12
Enable configuration management via viper

supreeth7:
### What type of PR is this?
Feature

- This PR leverages [viper](https://github.com/spf13/viper) to enable effective configuration management for Backplane CLI.
- The configuration proxy and Backplane URL set by the users via the `HTTPS_PROXY` and `BACKPLANE_URL` environment variables are respected and given preference over the values defined in the configuration file which wasn't the case before.

Resolves: [OSD-15922](https://issues.redhat.com/browse/OSD-15922)

- [x] Ran unit tests locally


wanghaoran1988:
```
{
	"proxy-url": "${URL}",
	"ocm-prod": "${URL}",
	"ocm-stg":"${URL}",
	"ocm-int":"${URL}",
	"bp-url-prod": "${URL}",
	"bp-url-stg":"${URL}",
	"bp-url-int":"${URL}"
}
```
This is a breaking change for the cli config,  I don't think it's a good option to put all those urls in this config.

wanghaoran1988:
> > I don't think it's a good option to put all those urls in this config
> 
> Thanks for taking a look, I understand, but what is stopping us from having a single config file? Is it a security concern? Anyhow we are exposing URLs in the current config, with this we don't have to switch between multiple config files until we have a mapping defined in OCM API.
Single config means you are more easy to make mistake, expose different env remind you which env you are accessing.

supreeth7:
The vulnerabilities outweigh the use case.
Let us have SDA incorporate this on the OCM side where we have a field in the provision shard object that is the `backplane_url` where SDA can populate that in the `environment` DB table exposing it in the cluster payload.

supreeth7:
@wanghaoran1988 I've removed the explicit mapping which doesn't break the CLI. This PR integrates viper for config management please take a look.
This work will be followed by https://issues.redhat.com/browse/OSD-15878

pkg/cli/config/config.go
```diff
@@ -1,35 +1,21 @@
...
 	"path/filepath"
 
 	"github.com/openshift/backplane-cli/pkg/info"
+	"github.com/spf13/viper"
 )
 
 type BackplaneConfiguration struct {
-	URL      string `json:"url"`
-	ProxyURL string `json:"proxy-url"`
+	URL      string `mapstructure:"url"`
```
wanghaoran1988, line 13: Could you help me understand why this?

supreeth7: From the [documentation](https://github.com/spf13/viper#unmarshaling):

> Viper uses [github.com/mitchellh/mapstructure](https://github.com/mitchellh/mapstructure) under the hood for unmarshaling values which uses mapstructure tags by default.

pkg/cli/config/config.go
```diff
@@ -40,51 +26,54 @@ func getDefaultConfigPath() (string, error) {
...
-			}
+		if err := viper.Unmarshal(&bpConfig); err != nil {
+			return bpConfig, err
+		}
+	}
 
-			return bpConfig, nil
+	// Check if user has explicitly defined backplane URL
+	// It has higher precedence over the config file
+	bpURL, hasURL := os.LookupEnv(info.BACKPLANE_URL_ENV_NAME)
```
wanghaoran1988, line 62: viper does support read configs considering environment setup, could you try to see how to integrate this? https://github.com/spf13/viper#working-with-environment-variables

wanghaoran1988: I see no value use viper without using this feature.

supreeth7: I had this added in another branch, moved it here, thanks

PR #81, opened 2023-04-21
Re-add build-ocm-container makefile target

hectorakemp:
### What type of PR is this?

feature

Re-adds the ability to run `make build-ocm-container` on top of `ocm-container`, allowing us to use backplane inside of `ocm-container` instances. 

To test: 
Run `make build-ocm-container`. It should now use this format for logging in: 
```
[17:38:23] hectorkemp:backplane-cli git:(readd-make-ocm-container) $ ocm-container a-cluster
Logging into cluster a-cluster
INFO[0001] Target cluster                                ID=123456 Name=a-cluster
```

_Resolves #_

- [ ] Ran unit tests locally

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,3 +118,10 @@ skopeo-push: image
 		--dest-creds "${QUAY_USER}:${QUAY_TOKEN}" \
 		"docker-daemon:${IMAGE_URI_LATEST}" \
 		"docker://${IMAGE_URI_LATEST}"
+
+build-ocm-container:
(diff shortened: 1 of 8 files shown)


mjlshen (commented):
Does this still need to live in the backplane-cli repo? I think it did in the past because it was a private repo and this was part of the hack to let ocm-container get this setup, but now that backplane-cli is a public repo I feel like this could all live within ocm-container itself

hectorakemp:
@mjlshen I agree - there's also a pull request that I did not see from @iamkirkbater (https://github.com/openshift/ocm-container/pull/174) which fixes this issue in `ocm-container`. I'll close this PR if no one has any objections.

PR #90, opened 2023-04-27
Improve error message clarity when the backplane url is empty

mjlshen:
### What type of PR is this?

cleanup

[OSD-16175](https://issues.redhat.com//browse/OSD-16175)

As SREs are switching to the new backplane-cli, sometimes they get a confusing error message:

```
2023/04/27 11:52:41 getting AWS credentials from backplane-api
2023/04/27 11:52:42 can't find backplane url: %!w(<nil>)
```

The message comes from the following block of code when there there is no error, but the format string attempts to insert one:

```go
		if err != nil || bpConfig.URL == "" {
			return fmt.Errorf("can't find backplane url: %w", err)
		}
```

This PR attempts to address it by returning this instead:
```go
		if bpConfig.URL == "" {
			return errors.New("empty backplane url - check your backplane-cli configuration")
		}
```

### DIFF
--- a/cmd/ocm-backplane/cloud/console.go
+++ b/cmd/ocm-backplane/cloud/console.go
@@ -3,6 +3,7 @@ package cloud
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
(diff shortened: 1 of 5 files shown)


supreeth7 (approved):
/lgtm
/approve

PR #91, opened 2023-04-28
Give faster feedback when logging into hibernating clusters

mjlshen:
### What type of PR is this?

cleanup

Currently, when we try to login to a cluster that is hibernating, it tries for 30 seconds before returning an error. This PR checks right away so we can give faster feedback since a login should never work for a hibernating cluster.

```
❯ time ocm backplane login ${HIBERNATING_CLUSTER}
INFO[0000] Target cluster                                ID=${ID} Name=${NAME}
ERRO[0030] cluster ${NAME} is hibernating, login failed 
ocm backplane login ${HIBERNATING_CLUSTER}  0.07s user 0.08s system 0% cpu 30.949 total
```

[OSD-16184](https://issues.redhat.com//browse/OSD-16184)

### DIFF
--- a/cmd/ocm-backplane/login/login.go
+++ b/cmd/ocm-backplane/login/login.go
@@ -99,7 +99,6 @@ func runLogin(cmd *cobra.Command, argv []string) (err error) {
 
 	// Get Backplane configuration
 	bpConfig, err := config.GetBackplaneConfiguration()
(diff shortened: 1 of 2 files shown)


supreeth7 (approved):
/lgtm
/approve

PR #105, opened 2023-05-15
OSD-16307: Add unit tests to session command

samanthajayasinghe:
### What type of PR is this?

Unit tests

### What this PR does / Why we need it?

Add tests to session command 

### Which Jira/Github issue(s) does this PR fix?

https://issues.redhat.com/browse/OSD-16307

### Pre-checks (if applicable)

- [x] Ran unit tests locally
- [x] Validated the changes in a cluster
- [ ] Included documentation changes with PR

PR #106, opened 2023-05-17
Config subcommand

supreeth7:
### What type of PR is this?
Feature

This PR adds a new `config` command to Backplane CLI making configuration management easy.

- `ocm backplane config get [url/proxy-url/all]` retrieves the Backplane CLI configuration variables.
- `ocm backplane config set url https://example.com` sets the Backplane CLI configuration.

Uses [viper](https://github.com/spf13/viper) in the backend.

Resolves: https://issues.redhat.com/browse/OSD-15878

Unit tests will be followed with another PR

- [x] Ran unit tests locally

### DIFF
--- a/cmd/ocm-backplane/config/config.go
+++ b/cmd/ocm-backplane/config/config.go
@@ -1,0 +1,29 @@
+package config
+
+import (
+	"github.com/spf13/cobra"
+)
+
+const (
+	ProxyURLConfigVar = "proxy-url"
+	URLConfigVar      = "url"
+)
+
+func NewConfigCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "config",
+		Short: "Get or set backplane-cli configuration",
+		Long: `Get or set backplane-cli configuration variables.
+The location of the configuration file is gleaned from ~/.config/backplane/config.json or the 'BACKPLANE_CONFIG' environment variable if set.
+
+The following variables are supported:
+url         Backplane API URL
+proxy-url   Squid proxy URL
+`,
+		SilenceUsage: true,
+	}
+
+	cmd.AddCommand(newGetCmd())
+	cmd.AddCommand(newSetCmd())
+	return cmd
+}
--- a/cmd/ocm-backplane/config/config.go
+++ b/cmd/ocm-backplane/config/config.go
@@ -7,6 +7,7 @@ import (
 const (
 	ProxyURLConfigVar = "proxy-url"
 	URLConfigVar      = "url"
+	SessionConfigVar  = "session-dir"
 )
 
 func NewConfigCmd() *cobra.Command {
@@ -19,6 +20,7 @@ The location of the configuration file is gleaned from ~/.config/backplane/confi
 The following variables are supported:
 url         Backplane API URL
 proxy-url   Squid proxy URL
+session-dir Backplane CLI session directory
 `,
 		SilenceUsage: true,
 	}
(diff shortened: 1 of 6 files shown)


cmd/ocm-backplane/config/set.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func setConfig(cmd *cobra.Command, args []string) error {
+	// Retrieve default Backplane CLI config path, $HOME/.config/backplane/config.json
+	configPath, err := config.GetConfigFilePath()
+	if err != nil {
+		return err
+	}
+
+	if configFlags.proxyURL == "" {
```
bmeng, line 50:
```
$ ocm backplane2 config set --url https://www.example.com
ERRO[0000] proxy-url should not be empty                
$ ocm backplane2 config set --proxy-url http://www.example.com:3128
ERRO[0000] url should not be empty                      
```

Should be able to set them separately?

supreeth7: We cannot unfortunately, `viper` overwrites the entire config file while saving, hence all the config parameters should be provided.

cmd/ocm-backplane/config/config.go
```diff
@@ -0,0 +1,29 @@
...
+
+func NewConfigCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "config",
+		Short: "Get or set backplane-cli configuration",
+		Long: `Get or set backplane-cli configuration variables.
+The location of the configuration file is gleaned from ~/.config/backplane/config.json or the 'BACKPLANE_CONFIG' environment variable if set.
+
+The following variables are supported:
+url         Backplane API URL
```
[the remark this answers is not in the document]
supreeth7: Variables are a direct mapping of the configuration variables.
In the config file, we call it `url` hence we ought to name the variable the same.
If we change it to `backplane-url` it will be a breaking change and users will have trouble after upgrading.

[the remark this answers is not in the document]
bmeng: yeah, What I exact mean is the name in the config file, so I mention it is out of scope here.

cmd/ocm-backplane/config/set.go
bmeng: How about we read the file when setting, if the file exists we can load the existing value and update the one from the parameter?

cmd/ocm-backplane/config/config.go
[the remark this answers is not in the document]
supreeth7: Understood, we could've done it earlier, but we're too deep down the road now, would break the existing functionality :)

cmd/ocm-backplane/config/set.go
bmeng: I am not sure which way is better approach, but I feel all the parameters should be set at the same time looks a bit inconvenient.
And I think with the other PR in the queue, there will be more variables added to the configfile. like this https://github.com/openshift/backplane-cli/pull/105/files#diff-66bf83773b78df966f05c4b0ea1985e249c5fc46ae55f942a0adcc431e904d3cR337

bmeng:
cc @samanthajayasinghe I think you will need the new variable in the configuration file can be get/set via this as well.

cmd/ocm-backplane/config/set.go
supreeth7: If the variable list is going to increase, I'll port to allow individual variables instead

supreeth7:
@samanthajayasinghe I've included the `session-dir` config var for better UX, it was be an empty string if not specified during config creation.

PR #113, opened 2023-05-24
OSD-14573-warn-user-for-old-cli

Tafhim:
### What type of PR is this?

feature

In order to help users identify version related issue faster, this PR adds a warning message when they attempt to login to a cluster.

[OSD-14573](https://issues.redhat.com//browse/OSD-14573)

- [x] Ran unit tests locally

### DIFF
--- a/cmd/ocm-backplane/login/login.go
+++ b/cmd/ocm-backplane/login/login.go
@@ -91,6 +91,8 @@ func init() {
 func runLogin(cmd *cobra.Command, argv []string) (err error) {
 	var clusterKey string
 
+	utils.CheckBackplaneVersion(cmd)
+
 	// Get The cluster ID
 	if len(argv) == 1 {
 		// if explicitly one cluster key given, use it to log in.
--- a/cmd/ocm-backplane/cloud/console.go
+++ b/cmd/ocm-backplane/cloud/console.go
@@ -81,6 +81,8 @@ func init() {
 func runConsole(cmd *cobra.Command, argv []string) (err error) {
 	var clusterKey string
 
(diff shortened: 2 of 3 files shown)


cmd/ocm-backplane/login/login.go
```diff
@@ -283,6 +287,34 @@ func getUsernameFromJWT(token string) string {
...
+	// Get the latest version from the GitHub API
+	latestVersion, err := git.GetLatestVersion(ctx)
+	if err != nil {
+		logger.WithField("Fetch error", err).Warn("Could not fetch latest version from GitHub")
+		return
+	}
+
+	// Check if the local version is already up-to-date
+	if latestVersion.TagName == info.Version {
+		logger.WithField("Current version", info.Version).Info("Already up-to-date")
```
bmeng, line 311: Should better put this as debug level instead of info?

cmd/ocm-backplane/login/login.go
```diff
@@ -283,6 +287,43 @@ func getUsernameFromJWT(token string) string {
 	return claim.(string)
 }
 
+// This function checks the backplane version and aims to only
```
supreeth7, line 290: Function comments usually start with the function name itself as per the standard convention
```suggestion
// checkBackplaneVersion checks the backplane version and aims to only
```

supreeth7 (changes requested):
Looks great overall.
This would be a good warning-level log for the initializing commands.
`login` being one, can you also add the `checkBackplaneVersion` function call to `cloud console` as well? Because users can view the cloud console without logging into the cluster.

supreeth7 (approved):

PR #114, opened 2023-05-24
Add commands to get an STS token for a role from an ocm token, and to assume roles through a chain of ARNs

AlexVulaj:
### What type of PR is this?

feature

Adds one command to fetch STS credentials for given role, and another to assume a chain of roles from a starting point to a cluster's account

[OSD-16489](https://issues.redhat.com//browse/OSD-16489)
[OSD-16490](https://issues.redhat.com//browse/OSD-16490)

Waiting on [OSD-16494](https://issues.redhat.com//browse/OSD-16494) to finish

- [x] Ran unit tests locally

To test locally:
In an AWS account, create a role with the trust policy in the attached ticket. Run this command with the arn for that role passed in as --role-arn, i.e. backplane cloud token --role-arn [role arn]. If successful, the command should output AWS STS Credentials as a JSON payload.

### DIFF
--- a/cmd/ocm-backplane/cloud/assume.go
+++ b/cmd/ocm-backplane/cloud/assume.go
@@ -1,0 +1,140 @@
+package cloud
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/credentials"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"github.com/openshift/backplane-cli/pkg/awsUtil"
+	"github.com/openshift/backplane-cli/pkg/cli/config"
+	"github.com/openshift/backplane-cli/pkg/utils"
+	"github.com/spf13/cobra"
+	"io"
+)
+
+const (
+	DefaultRoleArn = "arn:aws:iam::922711891673:role/SRE-Support-Role"
+)
+
+var assumeArgs struct {
+	roleArn string
+	output  string
+}
+
+var AssumeCmd = &cobra.Command{
+	Use:   "assume [CLUSTERID|EXTERNAL_ID|CLUSTER_NAME|CLUSTER_NAME_SEARCH]",
+	Short: "Performs the assume role chaining necessary to generate temporary access to the customer's AWS account",
+	Long: `Performs the assume role chaining necessary to generate temporary access to the customer's AWS account
+
+This command is the equivalent of running "aws sts assume-role-with-web-identity --role-arn [role-arn] --web-identity-token [ocm token] --role-session-name [email from OCM token]" behind the scenes,
+where the ocm token used is the result of running "ocm token". Then, the command makes a call to the backplane API to get the necessary jump roles for the cluster's account. It then calls the
+equivalent of "aws sts assume-role --role-arn [role-arn] --role-session-name [email from OCM token]" repeatedly for each role arn in the chain, using the previous role's credentials to assume the next
+role in the chain.
+
+This command will output sts credentials for the target role in the given cluster in formatted JSON. If no "role-arn" is provided, a default role will be used.
+`,
+	Example: `With default role:
+backplane cloud assume e3b2fdc5-d9a7-435e-8870-312689cfb29c -oenv
+
+With given role:
+backplane cloud assume e3b2fdc5-d9a7-435e-8870-312689cfb29c --role-arn arn:aws:iam::1234567890:role/read-only -oenv`,
+	Args: cobra.ExactArgs(1),
+	RunE: runAssume,
+}
+
+func init() {
+	flags := AssumeCmd.Flags()
+	flags.StringVar(&assumeArgs.roleArn, "role-arn", DefaultRoleArn, "The arn of the role for which to start the role assume process.")
+	flags.StringVarP(&assumeArgs.output, "output", "o", "env", "Format the output of the console response.")
+}
+
+type assumeChainResponse struct {
+	AssumptionSequence []namedRoleArn `json:"assumption_sequence"`
+}
+
+type namedRoleArn struct {
+	Name string `json:"name"`
+	Arn  string `json:"arn"`
+}
+
+func runAssume(_ *cobra.Command, args []string) error {
+	ocmToken, err := utils.DefaultOCMInterface.GetOCMAccessToken()
+	if err != nil {
+		return fmt.Errorf("failed to retrieve OCM token: %w", err)
+	}
+
+	bpConfig, err := config.GetBackplaneConfiguration()
+	if err != nil {
+		return fmt.Errorf("error retrieving backplane configuration: %w", err)
+	}
+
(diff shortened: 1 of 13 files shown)


cmd/ocm-backplane/cloud/assume.go
```diff
@@ -0,0 +1,146 @@
...
+		return fmt.Errorf("failed to get target cluster: %w", err)
+	}
+
+	backplaneClient, err := utils.DefaultClientUtils.MakeRawBackplaneAPIClientWithAccessToken(bpConfig.URL, *ocmToken)
+	if err != nil {
+		return fmt.Errorf("failed to create backplane client with access token: %w", err)
+	}
+
+	// TODO Change this to the actual function we'll be using
+	response, err := backplaneClient.GetBackplaneClusterClusterId(context.TODO(), clusterId)
```
geowa4, line 99: I believe this requires https://github.com/openshift/backplane-api/pull/11 to be merged and the code autogenerated.

pkg/utils/jwt.go
```diff
@@ -0,0 +1,33 @@
+package utils
+
+import (
+	"fmt"
+	"github.com/golang-jwt/jwt/v4"
+)
+
+func GetFieldFromJWT[T any](token string, field string) (T, error) {
```
bergmannf, line 8: Seems that the generic type is only ever string - so it feels that this could just as well be non-generic until needed?

AlexVulaj: I could go either way, since a function that only returns strings here would break if someone requested a differently typed field.

I think what I'll do is rename this to `GetStringFieldFromJWT` and remove the parameterized type. That way it's explicit.

pkg/utils/output.go
```diff
@@ -0,0 +1,32 @@
...
+	"encoding/json"
+	"fmt"
+	"sigs.k8s.io/yaml"
+)
+
+type Output interface {
+	EnvFormat() string
+}
+
+func RenderOutput[T Output](outputFormat string, data T) (string, error) {
```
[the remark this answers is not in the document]
bergmannf: It's not a big issue so if you are worried to break things I guess we can keep it in.
I just feel the generics make the signature more complex than it needs to be for no real generic-benefit as only a single branch even uses it.

cmd/ocm-backplane/cloud/assume.go
```diff
@@ -0,0 +1,146 @@
+package cloud
+
+import (
+	"context"
+	_ "embed"
```
geowa4, line 5: I think the import for "embed" can be removed.

pkg/utils/output.go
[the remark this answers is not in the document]
AlexVulaj: I could look into it. I personally don't love having "util" style packages depend on other packages (in this case, `utils` would need to depend on `awsUtil` but I'll see if it works out. It would be nice to use `awsUtil.AWSCredentialsResponse` instead of `T` here.

bmeng:
I assume this should be ready for review?

AlexVulaj:
@bmeng everything should be functionally complete - though I just resolved some merge conflicts and it looks like they've cause some issues. I'll dig into those.

AlexVulaj:
Build should be good now. I had to regenerate the mocks after the API updates.

cmd/ocm-backplane/cloud/assume.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"github.com/openshift/backplane-cli/pkg/awsUtil"
+	"github.com/openshift/backplane-cli/pkg/cli/config"
+	"github.com/openshift/backplane-cli/pkg/utils"
+	"github.com/spf13/cobra"
+	"io"
+)
+
+const (
+	DefaultRoleArn = "arn:aws:iam::922711891673:role/SRE-Support-Role"
```
bmeng, line 18: Is it a good idea to put our aws account id to public? 
Should be better if we read this as environment, and we can read it from the code?
Or we can also put this in to the config file, and we can use it directly. 
We just need to add another field to the config file https://github.com/openshift/backplane-cli/blob/main/cmd/ocm-backplane/config/config.go#L7

bmeng:
@AlexVulaj  Could you help update the README.md to include the usage for the two new sub-commands?
I was trying to execute them locally, but seems none of them working for me. With some weird errors.

Like:
```
ERRO[0006] failed to unmarshal response: invalid character 'p' after top-level value 
```

AlexVulaj:
@bmeng I can update the readme. To note - there should also be an example or two on each of the commands if you do a `-h`.

cmd/ocm-backplane/cloud/assume.go
AlexVulaj: I had the same concerns, but @jharrington22 reassured us that it was fine to expose the account ID since the role can't be assumed without an SSO account anyway.

https://redhat-internal.slack.com/archives/C058T1YEDPV/p1684870169269229?thread_ts=1684868454.749649&cid=C058T1YEDPV

cmd/ocm-backplane/cloud/assume.go
```diff
@@ -0,0 +1,140 @@
+package cloud
+
+import (
```
bmeng, line 3: A small one, could you sort a bit about the deps? Also in the other files as well.

cmd/ocm-backplane/cloud/token.go
```diff
@@ -0,0 +1,70 @@
...
+	bpConfig, err := config.GetBackplaneConfiguration()
+	if err != nil {
+		return fmt.Errorf("error retrieving backplane configuration: %w", err)
+	}
+	svc, err := awsUtil.StsClientWithProxy(bpConfig.ProxyURL)
+	if err != nil {
+		return fmt.Errorf("error creating STS client: %w", err)
+	}
+
+	result, err := awsUtil.AssumeRoleWithJWT(*ocmToken, tokenArgs.roleArn, svc)
```
bmeng, line 51: As we are going to assume role with the payer account in general, do we want to set it as the default value when people do not pass anything? Like the `assume` subcommand

cmd/ocm-backplane/cloud/assume.go
AlexVulaj: My IDE automatically sorts imports for me, and this is how it ended up. I also tried a `$ gofmt -w cmd/ocm-backplane/cloud/assume.go` and it left these imports as is. Is there something specific you're looking for here?

bmeng: ok, looks like there are different sort styles, my IDE uses `goimports` by default, I believe some of them are using `gofmt` by default. As we do not have a unified policy here. Let's leave it for now.

cmd/ocm-backplane/cloud/token.go
AlexVulaj: I could go either way on a default here. My understanding is that this command, "token", is going to be used for other flows at some point and having a default might not make sense. If you or anyone else feel strongly I'm happy to add in a default.

PR #124, opened 2023-06-13
Create config directory if it doesn't exist

cblecker:
### What type of PR is this?
bug

When setting a config option, if the directory doesn't exist you'll get an error. This ensures the directory is created. Will return an error if it can't create the directory.

_Resolves #_

- [ ] Ran unit tests locally


cmd/ocm-backplane/config/set.go
```diff
@@ -43,6 +44,13 @@ func setConfig(cmd *cobra.Command, args []string) error {
 		bpConfig.SessionDirectory = viper.GetString("session-dir")
 	}
 
+	// Create config directory if it doesn't exist
+	if dir, err := os.Stat(path.Dir(configPath)); os.IsNotExist(err) || !dir.IsDir() {
+		if err := os.Mkdir(path.Dir(configPath), 0750); err != nil {
```
geowa4, line 49: Should this be `MkDirAll` just in case?

cblecker: Thanks for asking that question. I had originally done this, but then thought to myself that what if you typo that directory.. do you want it creating a long tree because you typoed the early folder name if you set `BACKPLANE_CONFIG` explicitly.

The only way to know for sure the intent is to ask the user.. So now I do! And if they confirm, then I'm now using `MkdirAll`

geowa4: Awesome, looks good!

PR #125, opened 2023-06-15
fix the golangci yml name

bmeng:
### What type of PR is this?

_(bug)_

Update the golangci config file name, and enable some linters by default

_Resolves #_

- [ ] Ran unit tests locally


bmeng:
@mjlshen FYI, as the golangci.yml has incorrect name before, so the `gosec` linter was not run, and the test failed due to some of the failures. I will need to fix them firstly

PR #127, opened 2023-06-15
OSD-15841: Check proxy and API connection if login failed

samanthajayasinghe:
### What type of PR is this?

feature

Check the API connection via HTTP proxy transmission and provide a proper error message to the end user 

[OSD-15841](https://issues.redhat.com//browse/OSD-15841)

- [x] Ran unit tests locally
- [] Included documentation changes with PR

### DIFF
--- a/pkg/cli/config/config.go
+++ b/pkg/cli/config/config.go
@@ -1,8 +1,12 @@
 package config
 
 import (
+	"errors"
+	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/spf13/viper"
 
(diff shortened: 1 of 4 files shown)


pkg/cli/config/config.go
```diff
@@ -70,3 +74,46 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
...
 	return bpConfig, nil
 }
+
+// CheckAPIConnection validate API connection via configured proxy and VPN
+func (config BackplaneConfiguration) CheckAPIConnection() error {
+
+	// Check backplane Proxy URL
+	if config.ProxyURL == "" {
+		//Cannot connect to backplane API URL: Check if you need to use a proxy/VPN to access backplane
+		return errors.New("empty proxy url - check your backplane-cli configuration in $HOME/.config/backplane/config.json")
```
bmeng, line 84: The backplane config file could be other if it is set by env?

pkg/cli/config/config.go
```diff
@@ -70,3 +74,46 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
...
+
+	if !connectionOk {
+		return err
+	}
+
+	return nil
+}
+
+// pingBackplaneAPIviaProxy returns status of the the API connection
+func (config BackplaneConfiguration) pingBackplaneAPIviaProxy() (bool, error) {
```
bmeng, line 98: Can we rename the function name here? As the implementation is to send a normal http request, the `ping` here might be a bit confusing.

pkg/cli/config/config.go
samanthajayasinghe: Thanks @bmeng , Derive the path dynamically

pkg/cli/config/config.go
samanthajayasinghe: Rename the function to `testHttpRequestToBackplaneAPI()`, but let me know if you have a better option  :)

PR #143, opened 2023-07-07
Improve error handling for login connection issues

feichashao:
### What type of PR is this?

Bug

If login returns an authentication issues (4xx), and we don't have a proxy set, the current logic will show the error `cannot connect to backplane API URL, check if you need to use a proxy/VPN to access backplane`.

The PR changes the error handling to only print such error when the error returned from `doLogin` is exactly a network issue. And if the error returned from `doLogin` is a network issue, we don't need to test the connection again, thus removing the `bpConfig.CheckAPIConnection` check. 

https://issues.redhat.com/browse/OSD-17374

_Resolves #_

Tested locally.  
For connection issue, it will show   
```
./ocm-backplane login xxxxx
ERRO[0006] cannot connect to backplane API URL, check if you need to use a proxy/VPN to access backplane: unable to connect to backplane api: Post "https://localhost:8001/backplane/login/xxxx/": dial tcp [::1]:8001: connect: connection refused 
```

For authentication issue, it will show
```
./ocm-backplane login xxxxx
ERRO[0008] can't login to cluster: error from backplane: 
 Status Code: 403
 Message: no backplane role found for user yyyyy /FakeResource err/BackplaneOsdCeeResource false / 
```

- [ ] Ran unit tests locally

### DIFF
--- a/cmd/ocm-backplane/login/login.go
+++ b/cmd/ocm-backplane/login/login.go
@@ -51,7 +51,7 @@ var (
 		SilenceUsage: true,
 	}
 
-	ErrLoginConnection = errors.New("unable to connect to backplane api")
+	errLoginConnection = errors.New("unable to connect to backplane api")
 )
 
 func init() {
@@ -210,7 +210,7 @@ func runLogin(cmd *cobra.Command, argv []string) (err error) {
 			// Hibernating, print an error
 			return fmt.Errorf("cluster %s is hibernating, login failed", clusterKey)
 		}
-		if errors.Is(err, ErrLoginConnection) {
+		if errors.Is(err, errLoginConnection) {
 			return fmt.Errorf("cannot connect to backplane API URL, check if you need to use a proxy/VPN to access backplane: %v", err)
 		}
 
@@ -355,7 +355,7 @@ func doLogin(api, clusterId, accessToken string) (string, error) {
 		// trying to determine the error
 		errBody := err.Error()
 		if strings.Contains(errBody, "dial tcp") || strings.Contains(errBody, "i/o timeout") {
-			return "", fmt.Errorf("%w: %w", ErrLoginConnection, err)
+			return "", fmt.Errorf("%w: %w", errLoginConnection, err)
 		}
 
 		return "", err
--- a/cmd/ocm-backplane/login/login.go
+++ b/cmd/ocm-backplane/login/login.go
@@ -50,6 +50,8 @@ var (
 		RunE:         runLogin,
 		SilenceUsage: true,
 	}
+
+	ErrLoginConnection = errors.New("unable to connect to backplane api")
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


cmd/ocm-backplane/login/login.go
```diff
@@ -208,9 +210,7 @@ func runLogin(cmd *cobra.Command, argv []string) (err error) {
 			// Hibernating, print an error
 			return fmt.Errorf("cluster %s is hibernating, login failed", clusterKey)
 		}
-		// Check API connection with configured proxy
-		err = bpConfig.CheckAPIConnection()
```
samanthajayasinghe, line 212: Any reason to remove `bpConfig.CheckAPIConnection()`, which is ping to api end-point and validates the connection?
https://github.com/openshift/backplane-cli/blob/bcd5b3e43d6d7a8ec3f2654e9cd7e3fd8d4122de/pkg/cli/config/config.go#L92

cmd/ocm-backplane/login/login.go
```diff
@@ -50,6 +50,8 @@ var (
 		RunE:         runLogin,
 		SilenceUsage: true,
 	}
+
+	ErrLoginConnection = errors.New("unable to connect to backplane api")
```
samanthajayasinghe, line 54: maybe `errLoginConnection` since the variable is not use outside the login module

cmd/ocm-backplane/login/login.go
feichashao: I tired changing the logic to only perform `CheckAPIConnection` only when the error returned is `errLoginConnection` but fails the test cases. 

```
if errors.Is(err, errLoginConnection) {
			connectionErr := bpConfig.CheckAPIConnection()
			if connectionErr != nil {
				return fmt.Errorf("cannot connect to backplane API URL, check if you need to use a proxy/VPN to access backplane: %v", connectionErr)
			}
		}
```

Test result:
```
• Failure [1.378 seconds]
Login command
/Users/feichashao/git/github/backplane-cli/cmd/ocm-backplane/login/login_test.go:28
  check cluster login
  /Users/feichashao/git/github/backplane-cli/cmd/ocm-backplane/login/login_test.go:187
    should fail when BP API timeouts [It]
    /Users/feichashao/git/github/backplane-cli/cmd/ocm-backplane/login/login_test.go:294

    Expected
        <string>: can't login to cluster: unable to connect to backplane api: dial tcp i/o timeout
    to contain substring
        <string>: cannot connect to backplane API URL
```

while I was trying to tweak the test case, I think we can instead skip testing the api connectivity as the error message from `doLogin` has already did a "test" on the apiserver, we can expose the actual network error + a hint about proxy/vpn to make the error logic simpler.

samanthajayasinghe: I guess this [PR](https://github.com/openshift/backplane-cli/pull/150) fixed the error overwriting issue.

feichashao:
Closing this as https://github.com/openshift/backplane-cli/pull/151 can solve this problem.

PR #144, opened 2023-07-08
Add service and manager flags to backplane session

mrbarge:
### What type of PR is this?

Feature

This adds the same `--service` and `--manager` flags to `ocm backplane session` that are already in place for `ocm backplane login` in order to establish a backplane session to the cluster's service or management cluster respectively.

I use `backplane session` almost exclusively these days, but I miss these flags from `backplane login` quite a lot.

- [ ] Ran unit tests locally
- [X] Validated the changes in a cluster

### DIFF
--- a/cmd/ocm-backplane/session/session.go
+++ b/cmd/ocm-backplane/session/session.go
@@ -37,6 +37,22 @@ func NewCmdSession() *cobra.Command {
 			return validEnvs, cobra.ShellCompDirectiveNoFileComp
 		},
 	}
+
+	// Add login cmd specific flags
+	sessionCmd.Flags().BoolVar(
+		&options.Manager,
+		"manager",
+		false,
+		"Login to management cluster instead of the cluster itself.",
+	)
+
+	sessionCmd.Flags().BoolVar(
+		&options.Service,
+		"service",
+		false,
+		"Login to service cluster for the given hosted cluster or mgmt cluster.",
+	)
+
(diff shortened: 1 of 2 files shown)


pkg/cli/session/session.go
```diff
@@ -373,17 +396,17 @@ func (e *BackplaneSession) initClusterLogin(cmd *cobra.Command) error {
 		// Setting up the flags
 		err := login.LoginCmd.Flags().Set("multi", "true")
 		if err != nil {
-			return fmt.Errorf("error occered when setting multi flag %v", err)
+			return fmt.Errorf("error occurred when setting multi flag %v", err)
```
samanthajayasinghe, line 399: Good catch :)

cmd/ocm-backplane/session/session.go
```diff
@@ -37,6 +37,22 @@ func NewCmdSession() *cobra.Command {
...
+		"Login to management cluster instead of the cluster itself.",
+	)
+
+	sessionCmd.Flags().BoolVar(
+		&options.Service,
+		"service",
+		false,
+		"Login to service cluster for the given hosted cluster or mgmt cluster.",
+	)
+
```
samanthajayasinghe, line 55: next time, We should move these two flags to globalOpts and reuse them via `login` , `session` and `multi-login`

supreeth7 (approved):

PR #148, opened 2023-07-11
Refactor: Make manager & service flags global

supreeth7:
### What type of PR is this?
Feature

This PR makes the `manager` and `service` flags global so multiple commands can access it while maintaining flag consistency.

Please test this locally before you merge

- [x] Ran unit tests locally

### DIFF
--- a/pkg/cli/globalflags/globalflags.go
+++ b/pkg/cli/globalflags/globalflags.go
@@ -8,6 +8,8 @@ import (
 type GlobalOptions struct {
 	BackplaneURL string
 	ProxyURL     string
+	Manager      bool
+	Service      bool
 }
 
 func AddGlobalFlags(cmd *cobra.Command, opts *GlobalOptions) {
@@ -22,5 +24,17 @@ func AddGlobalFlags(cmd *cobra.Command, opts *GlobalOptions) {
 		"proxy",
 		"",
 		"URL of HTTPS proxy",
+	)
+	cmd.PersistentFlags().BoolVar(
+		&opts.Manager,
+		"manager",
+		false,
+		"Login to management cluster instead of the cluster itself.",
+	)
+	cmd.PersistentFlags().BoolVar(
+		&opts.Service,
+		"service",
+		false,
+		"Login to service cluster for the given hosted cluster or management cluster.",
 	)
 }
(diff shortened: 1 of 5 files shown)


pkg/cli/globalflags/globalflags.go
```diff
@@ -8,6 +8,8 @@ import (
 type GlobalOptions struct {
```
samanthajayasinghe, line 8: Nothing wrong here, but I feel that name does not align with `GlobalOptions` anymore since it contains `url`/`proxy`/`manager` & `service` . May be `GlobalLoginOptions` or `LoginOptions`

pkg/cli/globalflags/globalflags.go
```diff
@@ -8,6 +8,8 @@ import (
...
 	BackplaneURL string
 	ProxyURL     string
+	Manager      bool
+	Service      bool
 }
 
 func AddGlobalFlags(cmd *cobra.Command, opts *GlobalOptions) {
```
bmeng, line 15: This globalflags makes me a bit confusing. Is this something should be shared across all the subcommands?
But looks there is only one globalflag for backplane cli which is `verbosity` https://github.com/openshift/backplane-cli/blob/b84a15bb5dfe6ae3fa52b9a337fbdb40fdd26fa6/cmd/ocm-backplane/root.go#L63

supreeth7: These flags aren't meant to be shared across all subcommands. For example, `ocm backplane mangedjob create ...`, `ocm backplane scripts ...`, etc are meant to be confined to the cluster the user is logged into.

pkg/cli/globalflags/globalflags.go
supreeth7: We might add a few more flags on feature requests in the future, having a single data store with a more generic name (`GlobalOptions`) seems like a better option than confining to a specific use-case.

pkg/cli/globalflags/globalflags.go
bmeng: It sounds something we need to refactor to group the flags, and leave the global option real global

supreeth7: Thanks for merging, didn't get time to get back to this, we'll refactor global flags into flag groups soon.

PR #151, opened 2023-07-12
OSD-17374: Fixed backplane error overwriting issue when the proxy is not available

samanthajayasinghe:
### What type of PR is this?

_(bug/feature/cleanup/documentation)_

_Resolves #_

- [ ] Ran unit tests locally


feichashao:
/lgtm

The current logic makes sense. We see errors from login -> test the api connection e2e to determine the connection error.
The previous `CheckAPIConnection`  assumes the user must have a proxy, which causes confusions on the localhost cases where a proxy is not required.

PR #152, opened 2023-07-14
[OSD-16605] streaming logs for managedjob

bmeng:
### What type of PR is this?

_(feature)_

_Resolves #[OSD-16605](https://issues.redhat.com//browse/OSD-16605)_

- [ ] Ran unit tests locally

### DIFF
--- a/cmd/ocm-backplane/managedJob/createManagedJob.go
+++ b/cmd/ocm-backplane/managedJob/createManagedJob.go
@@ -3,14 +3,16 @@ package managedJob
 import (
 	"context"
 	"fmt"
+	"io"
 	"net/http"
+	"os"
 	"time"
 
-	"github.com/spf13/cobra"
-
 	BackplaneApi "github.com/openshift/backplane-api/pkg/client"
-	"github.com/openshift/backplane-cli/pkg/utils"
+	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/openshift/backplane-cli/pkg/utils"
 )
 
 var (
@@ -21,6 +23,7 @@ var (
 		clusterId     string
 		url           string
 		raw           bool
+		logs          bool
 	}
 )
 
@@ -49,6 +52,13 @@ func newCreateManagedJobCmd() *cobra.Command {
 		"w",
 		false,
 		"Wait until command execution is finished")
+
+	cmd.Flags().BoolVarP(
+		&options.logs,
+		"logs",
+		"",
+		false,
+		"Fetch logs from the pod for the running job")
 
 	return cmd
 }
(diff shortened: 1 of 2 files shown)


cmd/ocm-backplane/managedJob/createManagedJob.go
```diff
@@ -193,31 +213,94 @@ func waitForCreateJob(client BackplaneApi.ClientInterface, job *BackplaneApi.Job
...
+		return false, nil
+	})
+
+	return statusMessage, pollErr
+}
+
+// fetchJobLogs stream the log of the job to the console output when the job status is Running, Succeeded or Failed
+func fetchJobLogs(client BackplaneApi.ClientInterface, job *BackplaneApi.Job) error {
 
+	pollErr := wait.PollImmediate(5*time.Second, 1*time.Minute, func() (bool, error) {
```
samanthajayasinghe, line 241: I guess you need to increase time out to >1min, otherwise, this func will call 12 times and timeout.
I mean if the job is running for more than 1 min then this function will timeouts.

cmd/ocm-backplane/managedJob/createManagedJob.go
```diff
@@ -193,31 +213,94 @@ func waitForCreateJob(client BackplaneApi.ClientInterface, job *BackplaneApi.Job
...
 
-	return statusMessage, pollErr
+	if pollErr != nil {
+		return pollErr
+	}
+
+	version := "v2"
+	follow := true
+
+	resp, err := client.GetJobLogs(context.TODO(), options.clusterId, *job.JobId, &BackplaneApi.GetJobLogsParams{Version: &version, Follow: &follow})
```
samanthajayasinghe, line 276: In this case, logs are printed to the console when the job is finished. 
ideally, you can stream logs here 
```
case BackplaneApi.JobStatusStatusPending:
//Get the intermediate logs 
// append diff of the logs 
			return false, nil
```

cmd/ocm-backplane/managedJob/createManagedJob.go
bmeng: I think it should be waiting for 1 minute till the job status changes to Running/Completed/Failed ?

samanthajayasinghe: My understanding is `wait.PollImmediate`[1] calls the function each 5 sec and waits 1 minute to timeout. 
[1] https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollImmediate

I think you are right, if the status is not changed within 1 min then it may be related to other issues.

bmeng: yes, it is. It will wait up to 1 minute, but in the whole wait function, it just need to wait for the status to be expected.
The `GetJobLogs` is out of the wait function.

cmd/ocm-backplane/managedJob/createManagedJob.go
samanthajayasinghe: All good .. I tested the feature and logs show when the job status is `Running` |`Succeeded` | `Completed`

PR #170, opened 2023-08-04
Updated readme with instruction to users to use an updated version of go version

Dee-6777:
### What type of PR is this?

_documentation_

It instructs the user to use an updated go version so that there won't be any kube config files left after testing

_Resolves #[OSD-17636](https://issues.redhat.com/browse/OSD-17636)_

- [ ] Ran unit tests locally
- [X] Included documentation changes with PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ backplane-cli is a CLI tool to interact with [backplane api](https://github.com/
 
 ## Installation
 
+Go should be installed in your local system with version 1.19 
+
 ### Option 1: Download from release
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,6 +9,8 @@ backplane-cli is a CLI tool to interact with [backplane api](https://github.com/
 
 ## Installation
 
+Go should be installed in your local system with version >= 1.19
```
bmeng, line 12: I think maybe it is better to request the local go version to match with what the project is using.

Dee-6777: Ok, I'll keep it to only 1.19 then

PR #199, opened 2023-09-06
fix: add CI base image

supreeth7:
### What type of PR is this?
Feature

Adds a new file that indicates ProwCI which base image to consume giving the Backplane team more granular control over our CI; more specifically the go version it runs on.


supreeth7:
Scan is an optional check, it shouldn't block merging this PR.

PR #200, opened 2023-09-18
TestJob to use base64 for library import

feichashao:
### What type of PR is this?

bug

For testjob having library import, the current workflow will wrap the library file into a string, use a cat command to put the library content to lib.sh inline, then replace the `source /path/to/lib.sh` to `source ./lib.sh`

In some cases, the `cat << EOF` may cause strange issues if the library content has special characters or special formats. For example, the lib in https://github.com/openshift/managed-scripts/pull/137 may result in error `unbound variable`.

This PR encode the library file to base64, and decode the file inline, so that we don't need to worry about the format inside the library. 

We can use the existing `lib-sourcer` example for testing, the behavior remains unchanged.
```
[managed-scripts/scripts/examples/lib-sourcer] $ ocm.stg backplane testjob logs openshift-job-dev-p9jjt 
This is an imported library being used in a managed script.
```

- [x] Ran unit tests locally

### DIFF
--- a/cmd/ocm-backplane/testJob/createTestJob.go
+++ b/cmd/ocm-backplane/testJob/createTestJob.go
@@ -235,15 +235,8 @@ func createTestScriptFromFiles() (*backplaneApi.CreateTestScriptRunJSONRequestBo
 //
 // Inline into function before source definition of example.sh
 // #!/bin/bash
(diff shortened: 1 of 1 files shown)


hectorakemp (approved):
Nice, that's a really smart little fix :+1:

PR #210, opened 2023-09-25
Add dryRun flag for backplane-cli

xiaoyu74:
### What type of PR is this?

_feature_

_Resolves https://issues.redhat.com/browse/OSD-18387_

- Tested in my dev cluster with the dry-run flag and it works as expected to only show the JSON format of the test job without creating it.
- The dry-run feature has been implemented in the backplane-api, so this PR is just to add dryRun flag to call the dryRun API.


cmd/ocm-backplane/testJob/createTestJob.go
```diff
@@ -67,6 +67,13 @@ Example usage:
 		"Optional library file to be passed in (must live in managed-scripts/scripts directory)",
 	)
 
+	cmd.Flags().BoolP(
+		"dry-run",
+		"d",
+		false,
+		"Use this flag to perform a dry run, showing the YAML of the Job without creating it",
```
supreeth7, line 74:
```suggestion
		"Use this flag to perform a dry run, yeilds the YAML of the job without creating it",
```

supreeth7 (changes requested):

cmd/ocm-backplane/testJob/createTestJob.go
xiaoyu74: @supreeth7 - Thanks for your time to review, I will update accordingly.

bmeng:
I think we should better not print the job id if it is a dry run?
```
$ ocm backplane2 testjob create -d
...
...

TestId: openshift-job-1695719774

**Comment by xiaoyu74** [comment]
> I think we should better not print the job id if it is a dry run?
> 
> ```
> $ ocm backplane2 testjob create -d
> ...
> ...
> 
> TestId: openshift-job-1695719774
> ```

```
if err != nil {
    utils.WriteJson(w, http.StatusInternalServerError, &utils.ErrorResponse{
        StatusCode: http.StatusInternalServerError,
        Message: fmt.Sprintf("Failed running dryrun for job on the cluster, error: %s", err),
    })
    return
}
utils.WriteJson(w, http.StatusOK, jobs.TestJobResult{
    TestId:  jobObjects.Pod.Name,
    Message: buf.String(),
    Status:  string(corev1.PodPending), // Always starts with Pending
})
```
@bmeng  - Thanks for your review. I think it's just defined in the backplane API, if we think the `TestId` should not be printed, we can update in API code then but it's a separated issue from this PR.

bmeng:
I think we still can deal with it from https://github.com/openshift/backplane-cli/pull/210/files#diff-85bea5392ad379d93bf38a12b6a10a1b754877f9fd898d19ebaaf89c1a0f860fR184
But it is not a big deal, let's leave it as is.

PR #222, opened 2023-10-14
When specifying image on macOS, force linux/amd64 platform

cblecker:
### What type of PR is this?
bug

When using the `--image` flag on a non `linux/amd64` platform, the pull/run will fail due to a platform/architecture mismatch. This comes up in particular when running on an arm64 based mac with podman.

This forces the `--platform linux/amd64` flag on all podman image pulls. If you're running on a linux/amd64 platform, then this is a noop. But if you are using podman for mac, this will ensure you are always running the linux/amd64 version of the console image.

_Resolves #_

- [ ] Ran unit tests locally


supreeth7 (approved):
/lgtm

PR #229, opened 2023-10-17
[OSD-18058] Fix a reported bug about backplane elevate with interactive shell

xiaoyu74:
### What type of PR is this?

_bug_

Bug fixing for the elevate command with the interactive shell: https://issues.redhat.com/browse/OSD-18058


pkg/elevate/elevate.go
```diff
@@ -64,10 +64,10 @@ func RunElevate(argv []string) error {
 
 	elevateCmd := "oc " + strings.Join(argv[1:], " ")
 
-	shell := "/bin/bash"
-
-	if len(argv) > 3 {
-		shell = argv[4]
+	// Make the Default SHELL environment variable to /bin/bash
+	shell := os.Getenv("SHELL")
+	if shell == "" {
```
samanthajayasinghe, line 69: What if throws an error for the empty shell, or we need to check the validity of `/bin/bash` shell before making it as default shell.

pkg/elevate/elevate.go
```diff
@@ -61,12 +62,23 @@ func RunElevate(argv []string) error {
 
 	logger.Debug("Adding impersonation RBAC allow permissions to kubeconfig")
 
-	elevateCmd := argv[1:]
+	elevateCmd := "oc " + strings.Join(argv[1:], " ")
+
+	// Make the Default SHELL environment variable to /bin/bash
+	shell := os.Getenv("SHELL")
+	if shell == "" || !utils.IsValidShell(shell) {
```
samanthajayasinghe, line 69: can we add one more test case to check the empty shell scenario? 
```
#set empty env
os.Setenv("SHELL","")
# call  elevate cmd 
# check the error 
```

pkg/utils/util.go
```diff
@@ -250,3 +250,17 @@ func CheckBackplaneVersion(cmd *cobra.Command) {
...
+	IsValidShell(shellPath string) bool
+}
+
+type DefaultShellChecker struct{}
+
+// Helper function to check if a shell is valid
+func (checker DefaultShellChecker) IsValidShell(shellPath string) bool {
+	_, err := os.Stat(shellPath)
+	return err == nil
+}
```
samanthajayasinghe, line 266: Not a blocker.. 
great to see that func move to interface and mock via the tests :) 

I feel that it shouldn't be part of `pkg/util/util.go` may be `pkg/util/shell.go`

pkg/elevate/elevate_test.go
```diff
@@ -7,9 +7,20 @@ import (
...
+// MockShellChecker defines a mock implementation that can simulate different shell validation scenarios.
+type MockShellChecker struct {
+	IsValid bool // Determines whether the IsValidShell will return true or false.
+}
+
+// IsValidShell implements the ShellCheckerInterface.
+func (m MockShellChecker) IsValidShell(shellPath string) bool {
+	return m.IsValid
+}
+
```
samanthajayasinghe, line 23: If needed, You can generate the mock for an interface. 
https://github.com/openshift/backplane-cli/blob/711f0a6a76a8a0d5e08b0e27c771648d512222a3/Makefile#L109

PR #235, opened 2023-10-23
Implement retry for eventual consistency in IAM updates

AlexVulaj:
### What type of PR is this?
feature/bug fix

The retry mechanism built into the `sts.Client` from the AWS SDK doesn't properly handle issues of IAM updates requiring a few seconds to resolve, so we have to roll our own retry that recreates the client on every failure. See https://github.com/aws/aws-sdk-go-v2/issues/2332

[OSD-18992](https://issues.redhat.com//browse/OSD-18992)

- [x] Ran unit tests locally

### DIFF
--- a/cmd/ocm-backplane/cloud/assume.go
+++ b/cmd/ocm-backplane/cloud/assume.go
@@ -3,6 +3,7 @@ package cloud
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -16,15 +17,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const (
-	DefaultInitialRoleArn = "arn:aws:iam::922711891673:role/SRE-Support-Role"
-)
-
 var assumeArgs struct {
-	initialRoleArn string
-	output         string
-	debugFile      string
-	console        bool
+	output    string
+	debugFile string
+	console   bool
 }
 
 var StsClientWithProxy = awsutil.StsClientWithProxy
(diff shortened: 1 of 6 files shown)


cmd/ocm-backplane/cloud/assume.go
```diff
@@ -17,14 +18,20 @@ import (
...
 
 const (
-	DefaultInitialRoleArn = "arn:aws:iam::922711891673:role/SRE-Support-Role"
+	InitialRoleArnTemplate             = "arn:aws:iam::%v:role/SRE-Support-Role"
+	EnvProd                Environment = "prod"
+	ProdPayerAccountID                 = 922711891673
+	EnvStg                 Environment = "stg"
+	StgPayerAccountId                  = 277304166082
+	EnvInt                 Environment = "int"
+	IntPayerAccountId                  = 277304166082
```
samanthajayasinghe, line 27: What if moving the payerAccountID into the backplane config file [~/.config/backplane] ?
Otherwise, this exposes the payer accountID in a public repo for all envs .
There may be more supported env soon, [eg: `fedramp`], and this approach may need to change then.

cmd/ocm-backplane/cloud/assume.go
```diff
@@ -186,3 +208,23 @@ func runAssume(_ *cobra.Command, args []string) error {
...
+
+type Environment string
+
+func (e *Environment) String() string {
+	return string(*e)
+}
+
+func (e *Environment) Set(env string) error {
+	switch strings.ToLower(env) {
+	case "int", "stg", "prod":
```
samanthajayasinghe, line 220: This may not always true

cmd/ocm-backplane/cloud/assume.go
AlexVulaj: I spoke about this with @jharrington22 way back when we began this work and we determined that [the ARNs aren't secret so there's no issue with having them in GitHub](https://redhat-internal.slack.com/archives/C058T1YEDPV/p1684869351960399?thread_ts=1684868454.749649&channel=C058T1YEDPV&message_ts=1684869351.960399).

Would you still like me to move these to a config file?

cmd/ocm-backplane/cloud/assume.go
AlexVulaj: That's fair. Though we can change this as needed.

AlexVulaj:
Putting a hold on this momentarily.

For SRE convenience, we're going to remove the "env" flag entirely and have it pulled from the backplane config.

More context here: https://redhat-internal.slack.com/archives/CFJD1NZFT/p1698096531994639?thread_ts=1698096531.994639&cid=CFJD1NZFT

AlexVulaj:
/unhold

Moved the initial assume ARN to a config field.

Scutua (approved):

PR #236, opened 2023-10-25
[OSD-19019] Enable bp-cli to consume warning header from bp API's response

xiaoyu74:
### What type of PR is this?

_feature_

### What does this PR do / Why we need it?
To have bp-cli enabled to consume the warning header from bp-API's response, so that, when running managed-script with `ocm-backplane`, it can post warnings if the `clusterVersions` is not specified in the metadata.yaml but still allow the script continue execute.

It's interdependent with the [MR-303](https://gitlab.cee.redhat.com/service/backplane-api/-/merge_requests/303)

### DIFF
--- a/cmd/ocm-backplane/managedJob/createManagedJob.go
+++ b/cmd/ocm-backplane/managedJob/createManagedJob.go
@@ -175,7 +175,7 @@ func createJob(client BackplaneApi.ClientInterface) (*BackplaneApi.Job, error) {
 	// create job request
 	createJob := BackplaneApi.CreateJobJSONRequestBody{
 		CanonicalName: &options.canonicalName,
-		Parameters: &jobParams,
+		Parameters:    &jobParams,
 	}
 
 	// call create end point
@@ -183,6 +183,11 @@ func createJob(client BackplaneApi.ClientInterface) (*BackplaneApi.Job, error) {
 
 	if err != nil {
 		return nil, err
+	}
+
+	// Check for the warning header and display it if found.
+	if warningMsg := resp.Header.Get("Backplane-Warning"); warningMsg != "" {
+		fmt.Fprintln(os.Stderr, warningMsg)
 	}
 
 	if resp.StatusCode != http.StatusOK {
--- a/cmd/ocm-backplane/managedJob/createManagedJob.go
+++ b/cmd/ocm-backplane/managedJob/createManagedJob.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	BackplaneApi "github.com/openshift/backplane-api/pkg/client"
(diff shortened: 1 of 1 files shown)


samanthajayasinghe:
pls add a unit test to cover the new changes

cmd/ocm-backplane/managedJob/createManagedJob.go
```diff
@@ -175,11 +175,13 @@ func createJob(client BackplaneApi.ClientInterface) (*BackplaneApi.Job, error) {
...
-		Parameters: &jobParams,
+		Parameters:    &jobParams,
 	}
 
-	// call create end point
-	resp, err := client.CreateJob(context.TODO(), options.clusterID, createJob)
+	// call create endpoint and handle Backplane-Warning header
+	resp, err := utils.HandleBackplaneWarningHeader(func() (*http.Response, error) {
+		return client.CreateJob(context.TODO(), options.clusterID, createJob)
+	})
```
samanthajayasinghe, line 184: Do we need a util function to handle it ? What if just get a warning msg and log it as a warning

```
resp, err := client.CreateJob(context.TODO(), options.clusterID, createJob)
if warningMsg := resp.Header.Get("Backplane-Warning"); warningMsg != ""  {
   logger.Warnf("Backplane-Warning: %s", warningMsg)
}
```

xiaoyu74: @samanthajayasinghe - Sorry, I am bit confused. As mentioned in your first review comment, I got 2 points: 

- What if use the logger to print warnings **based on debug level**
- It is better to handle it via **commonplace** and apply to all subcommands.

My latest commit just trying to match the suggestions above: 

- HandleBackplaneWarningHeader function checks for the Backplane-Warning header in the response, and logs it **at different levels based on the current log level** (You also provided `jobs.go` which defined the flag for log level)
- I am trying to make the function **HandleBackplaneWarningHeader** reusable for any HTTP request. As long as other parts of cli use this function to handle HTTP responses, the warning logging behavior will be consistent across all backplane API calls.

I think we should have a quick sync directly to make sure I am not over thinking anything.

xiaoyu74:
> pls add a unit test to cover the new changes
Since the MR for validating clusterversion in backplane-api has been merged, can we merge this interdependent [PR-236](https://github.com/openshift/backplane-cli/pull/236) (which just to consume a custom warning header I defined in the API to log/print warning messages when end-user doesn't specify clusterVersions for managed-script in metadata.yaml) first.

We have a separate task https://issues.redhat.com/browse/OSD-19020 for writing test cases and I will add test case for bp-cli part accordingly.

PR #237, opened 2023-10-26
Move all functionality for assume command into credentials/console

AlexVulaj:
### What type of PR is this?
cleanup/feature

takes the logic from the new assume/token commands (which are not yet in use anywhere) and consolidates them into the existing commands. this will prevent SREs from having to change anything about their existing workflows.

[OSD-19436](https://issues.redhat.com//browse/OSD-19436)

- [x] Ran unit tests locally

### Steps to test locally
As a pre-requisite, run either `backplane cloud console` or `backplane cloud credentials` against an existing, non-ROSA cluster and make sure it still works as it always has. For the new flow:

1. Create a new ROSA STS cluster in stage. The "new flow" should be enabled in the Red Hat org in stage.
2. Once the cluster is created and has been set up with all support roles, use "backplane cloud console" or "backplane cloud credentials" against that cluster.
Note - at this point, it is not uncommon to see a message on the screen saying "Waiting for IAM policy changes to resolve..."
3. Verify that legitimate credentials or a console url are printed out.


AlexVulaj:
/hold

I'm putting this up now for review, but we must hold off on merging until [this conversation](https://redhat-internal.slack.com/archives/C058T1YEDPV/p1698348483360429?thread_ts=1698263212.125159&cid=C058T1YEDPV) is resolved. In short - the OCM endpoint that determines a cluster's support jump role needs an update to how it makes it determination.

cmd/ocm-backplane/cloud/console.go
```diff
@@ -132,13 +134,39 @@ func runConsole(cmd *cobra.Command, argv []string) (err error) {
...
+	}
+
+	var consoleResponse *ConsoleResponse
+	if supportRoleArn.Resource == OldFlowSupportRole { // cluster is using old backplane flow
+		consoleResponse, err = getCloudConsole(bpURL, clusterID)
+		if err != nil {
+			return err
+		}
+	} else { // cluster is using new backplane flow
+		targetCredentials, err := getNewFlowCredential(clusterID)
```
mrWinston, line 153: Minor Nitpick: Could we rename the function to just `getCloudCredentials` or something? I feel like the name `NewFlow` will get *old* pretty quickly :)

pkg/credentials/aws.go
```diff
@@ -0,0 +1,33 @@
+package credentials
+
+import "fmt"
+
+const (
+	// AwsCredentialsStringFormat format strings for printing AWS credentials as a string or as environment variables
+	AwsCredentialsStringFormat = `Temporary Credentials:
```
mrWinston, line 7: Really Minor Nitpick: Maybe use `\t` to indent the lines in the String. Just from looking at it in the code it's a bit confusing right now as it's the same indentation level as the surrounding code.

pkg/credentials/aws.go
```diff
@@ -0,0 +1,33 @@
+package credentials
```
mrWinston, line 1: Question: do you think it made sense to add some tests to this again?

pkg/utils/ocmWrapper.go
```diff
@@ -261,6 +262,18 @@ func (*DefaultOCMInterfaceImpl) IsProduction() (bool, error) {
 	return connection.URL() == "https://api.openshift.com", nil
 }
 
+func (*DefaultOCMInterfaceImpl) GetStsSupportJumpRole(clusterID string) (string, error) {
```
mrWinston, line 265: Minor Nitpick: Consider renaming this function to `GetStsSupportJumpRoleARN`.

mrWinston (commented):
Nice! I really like that this doesn't only make it easier for SREs but also makes the code easier to understand and reason about! I have yet to test the changes locally, which is why i didn't approve the PR right now. The comments I left don't neccessarily need to be addressed as they're all pretty minor.

cmd/ocm-backplane/cloud/console.go
AlexVulaj: I agree with this, and I struggled to find a good name because `getCloudCredential` already exists in this package in `credentials.go`. Any other ideas on a name?

pkg/credentials/aws.go
AlexVulaj: This was all copy-paste from what used to be in the old `console` and `credentials` commands, so I wanted to keep that identical.

I also don't believe these should be indented. Right now this will all print to a terminal as such:
```
Temporary Credentials:
AccessKeyID: test-key
SecretAccessKey: test-secret
SessionToken: test-token
Region: test-region
Expires: test-expiration
```

I believe that's what we'd want.

pkg/utils/ocmWrapper.go
AlexVulaj: My original thought was to go with the name of the OCM call itself, which is "StsSupportJumpRole", but I think specifying that it's the "ARN" is a good call here. I'll rename it.

pkg/credentials/aws.go
AlexVulaj: For functions like this I could go either way - let's err on the side of caution and add some tests.

AlexVulaj: For functions like this, it feels weird to test because my tests would essentially just do exactly what the function is already doing. For example...

```
	type fields struct {
		AccessKeyID     string
		SecretAccessKey string
		SessionToken    string
		Region          string
		Expiration      string
	}
	tests := []struct {
		name   string
		fields fields
		want   string
	}{
		{
			name:   "Contains no values",
			fields: fields{},
			want:   fmt.Sprintf(AwsExportFormat, "", "", "", ""), <-- this is just what the function itself is doing
		},
	}
```
That said - I could still go either way. Any thoughts?

cmd/ocm-backplane/cloud/console.go
AlexVulaj: Thought of a better name - coming in a commit.

pkg/credentials/aws.go
feichashao: It seems we don't have a good way to make indentation look nice in raw string.  https://github.com/golang/go/issues/32590

Let's keep it as is, as long as the output is what we want.

cmd/ocm-backplane/cloud/credentials_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	bpCredentials "github.com/openshift/backplane-cli/pkg/credentials"
```
feichashao, line 8: Suggest moving to line 25 to make backplane-cli import in the same group.

cmd/ocm-backplane/cloud/common.go
```diff
@@ -0,0 +1,128 @@
...
+	backplaneClient, err := utils.DefaultClientUtils.MakeRawBackplaneAPIClientWithAccessToken(bpConfig.URL, *ocmToken)
+	if err != nil {
+		return aws.Credentials{}, fmt.Errorf("failed to create backplane client with access token: %w", err)
+	}
+
+	response, err := backplaneClient.GetAssumeRoleSequence(context.TODO(), clusterID)
+	if err != nil {
+		return aws.Credentials{}, fmt.Errorf("failed to fetch arn sequence: %w", err)
+	}
+	if response.StatusCode != 200 {
```
supreeth7, line 77: Let's use HTTP status code from the package instead.

```suggestion
	if response.StatusCode != http.StatusOK {
```

cmd/ocm-backplane/cloud/console.go
```diff
@@ -30,14 +31,14 @@ type ConsoleResponse struct {
...
-var consoleStrFmt string = `Console Link:
+var consoleStrFmt = `Console Link:
   Link: %s`
 
 func (r *ConsoleResponse) String() string {
 	return fmt.Sprintf(consoleStrFmt, r.ConsoleLink)
 }
 
-// Environment variable that indicates if open by browser is set as default
+// EnvBrowserDefault Environment variable that indicates if open by browser is set as default
```
supreeth7, line 41:
```suggestion
// EnvBrowserDefault environment variable that indicates if open by browser is set as default
```

feichashao:
Hi @AlexVulaj , thanks for the PR to consolidate commands!  I am new to the "new flow", and I wonder if there's any steps that I can try that locally?

pkg/credentials/aws.go
iamkirkbater: Non-blocking - I wonder if we could use go's text/templates here to better define these strings separately.

cmd/ocm-backplane/cloud/common.go
```diff
@@ -0,0 +1,128 @@
...
+type assumeChainResponse struct {
+	AssumptionSequence []namedRoleArn `json:"assumptionSequence"`
+}
+
+type namedRoleArn struct {
+	Name string `json:"name"`
+	Arn  string `json:"arn"`
+}
+
+func getIsolatedCredential(clusterID string) (aws.Credentials, error) {
```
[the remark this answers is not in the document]
AlexVulaj: I went with "credential" because in `credentials.go` line 148 of this PR, the existing function for the "old flow" was named "getCloudCredential". Would you still like me to change this?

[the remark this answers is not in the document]
iamkirkbater: I'd be +1 to this change. There is no historical reason I can remember other than most likely me making a typo and that persisting forever :)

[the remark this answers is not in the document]
AlexVulaj: Done, thanks!

cmd/ocm-backplane/cloud/credentials_test.go
AlexVulaj: Good call, done

AlexVulaj:
Hey @feichashao , more details on the new flow can be found in the [ADR](https://docs.google.com/document/d/1IyJn8jCpbJNNPkx7hhaYfDy6yAf_NIf86ngrqsicK5g/edit#heading=h.hzsa87ps5uhr), and [this design doc](https://docs.google.com/document/d/1wo_onHDrbgk9hsJdL3xzJndJGEpOgjiTDidv-AbFniE/edit).

For steps to test locally, I'll add those to the description of this PR directly.

cmd/ocm-backplane/cloud/credentials.go
```diff
@@ -175,65 +110,38 @@ func runCredentials(cmd *cobra.Command, argv []string) error {
...
 		}
-		fmt.Println(creds)
-		return nil
-	default:
-		return fmt.Errorf("unsupported cloud provider: %s", cloudProvider)
-	}
-}
-
-// GetAWSV2Config allows consumers to get an aws-sdk-go-v2 Config to programmatically access the AWS API
-func GetAWSV2Config(backplaneURL string, clusterID string) (aws.Config, error) {
```
mjlshen, line 214: This function should not have been removed, osdctl depends on it https://github.com/openshift/osdctl/blob/04c8575e83b30d0199ff0d1cfd5b81761f870556/pkg/osdCloud/aws.go#L189

PR #243, opened 2023-11-01
Make sure region is always present for credentials and console links

AlexVulaj:
### What type of PR is this?

_bug_

Changes for the new flow made it so that "region" wasn't being set when requesting credentials for old and new clusters, or a console link for new clusters.

_Resolves OSD-19492_

- [x] Ran unit tests locally

### DIFF
--- a/pkg/awsutil/sts.go
+++ b/pkg/awsutil/sts.go
@@ -18,9 +18,9 @@ import (
 )
 
 const (
-	AwsFederatedSigninEndpoint = "https://signin.aws.amazon.com/federation"
-	AwsConsoleURL              = "https://console.aws.amazon.com/"
-	DefaultIssuer              = "Red Hat SRE"
+	AwsFederatedSigninEndpointTemplate = "https://%v.signin.aws.amazon.com/federation"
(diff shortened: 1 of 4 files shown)


pkg/awsutil/sts.go
```diff
@@ -18,9 +18,9 @@ import (
 )
 
 const (
-	AwsFederatedSigninEndpoint = "https://signin.aws.amazon.com/federation"
-	AwsConsoleURL              = "https://console.aws.amazon.com/"
-	DefaultIssuer              = "Red Hat SRE"
+	AwsFederatedSigninEndpointTemplate = "https://%v.signin.aws.amazon.com/federation"
+	AwsConsoleURL                      = "https://console.aws.amazon.com/"
```
mjlshen, line 22: This PR isn't adding these URLs, but why do these exist in backplane-cli now? These won't work in general for different partitions (e.g. GovCloud)

AlexVulaj: Backplane API used to handle generating this url for clusters using the old flow. For clusters using isolated backplane flow, that isn't the case. So the CLI is responsible in those instances.

PR #267, opened 2023-11-22
feat: make proxy configurable

typeid:
### What type of PR is this?

cleanup

We need the proxy to be configurable for programmatic access. Beyond that, this PR also allows passing the whole necessary configuration to create an AWSV2Config, instead of picking up config chunks through ENVs and external config files. 

Needed for https://issues.redhat.com/browse/OSD-19388\

_Resolves #_

- [x] Ran unit tests locally

**After these changes I tested the following:**
Old flow:
- ocm backplane cloud credentials
- ocm backplane cloud console

Non STS clusters:
- ocm backplane cloud credentials
- ocm backplane cloud console

New flow:
- ocm backplane cloud credentials
- ocm backplane cloud console

### DIFF
--- a/cmd/ocm-backplane/cloud/common.go
+++ b/cmd/ocm-backplane/cloud/common.go
@@ -14,9 +14,12 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/sts"
 	ocmsdk "github.com/openshift-online/ocm-sdk-go"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	BackplaneApi "github.com/openshift/backplane-api/pkg/client"
 	"github.com/openshift/backplane-cli/pkg/awsutil"
 	"github.com/openshift/backplane-cli/pkg/cli/config"
+	bpCredentials "github.com/openshift/backplane-cli/pkg/credentials"
 	"github.com/openshift/backplane-cli/pkg/utils"
+	logger "github.com/sirupsen/logrus"
 )
 
 const OldFlowSupportRole = "role/RH-Technical-Support-Access"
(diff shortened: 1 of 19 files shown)


cmd/ocm-backplane/cloud/common.go
```diff
@@ -12,18 +12,26 @@ import (
...
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/openshift/backplane-cli/pkg/awsutil"
+	"github.com/openshift/backplane-cli/pkg/cli/config"
 	"github.com/openshift/backplane-cli/pkg/utils"
 )
 
 const OldFlowSupportRole = "role/RH-Technical-Support-Access"
 
-var StsClientWithProxy = awsutil.StsClientWithProxy
+var StsClientWithProxy = awsutil.StsClient
```
AlexVulaj, line 24: I'd change the name of the var here too

cmd/ocm-backplane/cloud/common.go
```diff
@@ -12,18 +12,26 @@ import (
...
 var AssumeRoleWithJWT = awsutil.AssumeRoleWithJWT
 var NewStaticCredentialsProvider = credentials.NewStaticCredentialsProvider
 var AssumeRoleSequence = awsutil.AssumeRoleSequence
 
+// Wrapper for the configuration needed for cloud requests
+type CloudQueryConfig struct {
+	config.BackplaneConfiguration
+	OcmConnection *ocmsdk.Connection
```
mjlshen, line 32: I think `OcmConnection` should be part of `config.BackplaneConfiguration` - I don't think backplane works without OCM in general (not just `ocm-backplane cloud`)

Also minor nit on the naming, if I were importing this I'd end up making a `cloud.CloudQueryConfig` or maybe I'd alias it to `backplaneCloud.CloudQueryConfig` which doesn't make too much sense/a bit too much stutter for my liking :D

Could we rename this struct to just `Config`?

cmd/ocm-backplane/cloud/credentials.go
```diff
@@ -129,8 +132,13 @@ func runCredentials(cmd *cobra.Command, argv []string) error {
 }
 
 // getCloudCredentials returns Cloud Credentials Response
-func getCloudCredentials(backplaneURL string, ocmToken *string, cluster *cmv1.Cluster) (bpCredentials.Response, error) {
-	isolatedBackplane, err := isIsolatedBackplaneAccess(cluster)
+func getCloudCredentials(queryConfig *CloudQueryConfig, cluster *cmv1.Cluster) (bpCredentials.Response, error) {
```
mjlshen, line 135:
```suggestion
func (cfg *CloudQueryConfig) getCloudCredentials(cluster *cmv1.Cluster) (bpCredentials.Response, error) {
```

cmd/ocm-backplane/cloud/common.go
```diff
@@ -33,7 +41,7 @@ type namedRoleArn struct {
 	Arn  string `json:"arn"`
 }
 
-func getIsolatedCredentials(clusterID string, ocmToken *string) (aws.Credentials, error) {
+func getIsolatedCredentials(clusterID string, queryConfig *CloudQueryConfig, ocmToken *string) (aws.Credentials, error) {
```
mjlshen, line 44:
```suggestion
func (cfg *CloudQueryConfig) getIsolatedCredentials(clusterID string, ocmToken *string) (aws.Credentials, error) {
```

Depending on previous feedback, `clusterID string` would also become a member of `cfg`, same for `isIsolatedBackplaneAccess()`

cmd/ocm-backplane/cloud/common.go
mjlshen: Beyond that, imo the additional struct field that's useful for the `ocm-backplane cloud` functions is either a `clusterID string` or a `cluster *cmv1.Cluster`. If a caller can assemble all the necessary backplane config file fields + OCM connection + cluster ID, then I think that's the minimum ticket needed for entry.

pkg/cli/config/config.go
```diff
@@ -70,10 +70,15 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
...
 	bpConfig.URL = viper.GetString("url")
-	bpConfig.ProxyURL = viper.GetString("proxy-url")
 	bpConfig.SessionDirectory = viper.GetString("session-dir")
 	bpConfig.AssumeInitialArn = viper.GetString("assume-initial-arn")
 
+	// proxyURL is optional
+	proxyURL := viper.GetString("proxy-url")
+	if proxyURL != "" {
+		bpConfig.ProxyURL = &proxyURL
+	}
```
mjlshen, line 80: We want to enable programmatically using backplane without a proxy, but imo we shouldn't allow users of the CLI to bypass. It's a security requirement to use the proxy as a human.

mjlshen (changes requested):

cmd/ocm-backplane/cloud/common.go
typeid: Good catch, thanks :)

cmd/ocm-backplane/cloud/common.go
typeid: > I think OcmConnection should be part of config.BackplaneConfiguration - I don't think backplane works without OCM in general (not just ocm-backplane cloud)
This is something I wanted to do, but noticed it drags too many changes in this PR. If you don't mind, I'd create a card to do this as followup (e.g. we'd have to change the whole aws interface as well then).

> Also minor nit on the naming, if I were importing this I'd end up making a cloud.CloudQueryConfig or maybe I'd alias it to backplaneCloud.CloudQueryConfig which doesn't make too much sense/a bit too much stutter for my liking :D
Could we rename this struct to just Config?

Renamed to QueryConfig. If we want to move the `QueryConfig` struct into the global backplane config in the near future, I think we should keep the clusterid out of it as we will likely have to undo that work later.

pkg/cli/config/config.go
typeid: I think EDA will use backplane-cli directly and will need to run without a proxy. Are you fine with adding a warning in case the proxy is empty? The control layer should be the api anyway.

typeid:
Removed WIP label. Tested that this works programatically and tested that I can still create console/credentials for clusters.

pkg/cli/config/config.go
mjlshen: Yeah, enforcing this at the API is better - is it enforced there currently?

typeid: It is enforced on API level in the sense that the backplane-api is not accessible outside of the whitelisted IPs (proxy, CAD...)

mjlshen: Thanks, makes sense. As long as it remains impossible to "misconfigure" the config locally so that a user to use this without specifying a proxy url, I am ok.

cmd/ocm-backplane/cloud/common.go
```diff
@@ -96,16 +99,16 @@ func getIsolatedCredentials(clusterID string, ocmToken *string) (aws.Credentials
...
-	targetCredentials, err := AssumeRoleSequence(email, seedClient, roleAssumeSequence, bpConfig.ProxyURL, awsutil.DefaultSTSClientProviderFunc)
+	targetCredentials, err := AssumeRoleSequence(email, seedClient, roleAssumeSequence, cfg.BackplaneConfiguration.ProxyURL, awsutil.DefaultSTSClientProviderFunc)
 	if err != nil {
 		return aws.Credentials{}, fmt.Errorf("failed to assume role sequence: %w", err)
 	}
 	return targetCredentials, nil
 }
 
-func isIsolatedBackplaneAccess(cluster *cmv1.Cluster) (bool, error) {
+func isIsolatedBackplaneAccess(cluster *cmv1.Cluster, ocmConnection *ocmsdk.Connection) (bool, error) {
```
mjlshen, line 109:
```suggestion
func (cfg *QueryConfig) isIsolatedBackplaneAccess(cluster *cmv1.Cluster) (bool, error) {
```

Any reason to not make this (and kind of all the other functions) methods off of `QueryConfig`? To me, that's one of the main benefits of defining a `QueryConfig` in the first place - you don't have to worry about passing the right arguments?

I could be convinced otherwise though, currently I'm thinking as long as the public/external helper functions are methods that's good enough for me (e.g. skip this one and refactor `GetAWSV2Config` to be a method).

typeid: Agree that it makes sense, I was being a bit too careful with refactoring I suppose.

The following are now members of `QueryConfig` (and have been moved to `common.go`: 
- `getCloudCredentials` -> public 
- `getCloudCredentialsFromBackplaneAPI`
- `getCloudConsole` -> public

I also refactored `GetCloudConsole` the same way `GetCloudCredentials` has been refactored to enable calling it from `console.go`.

**After these changes I tested the following:**
Old flow:
- ocm backplane cloud credentials
- ocm backplane cloud console

Non STS clusters:
- ocm backplane cloud credentials
- ocm backplane cloud console

New flow:
- ocm backplane cloud credentials
- ocm backplane cloud console

mjlshen (approved):
/lgtm

Thanks!

PR #268, opened 2023-11-24
fix: use logger for iam changes log message

typeid:
### What type of PR is this?

bug

The `fmt.Println` log message per default interferes with the `-o env`. This PR changes the `Resolving...` output to use the logger which won't affect the below:

```
export $(ocm backplane cloud credentials $CLUSTER_ID -o env)
bash: export: `resolve...': not a valid identifier
```

_Resolves #_

- [ ] Ran unit tests locally


supreeth7 (approved):
/lgtm

PR #274, opened 2023-12-01
print the related namespace when login via manager cluster

bmeng:
### What type of PR is this?

_(feature)_

It prints the cluster related namespaces when login with the `--manager` option, since user may like to investigate the resources within

_Resolves #OSD-19081_

- [x] Ran unit tests locally

### DIFF
--- a/cmd/ocm-backplane/login/login.go
+++ b/cmd/ocm-backplane/login/login.go
@@ -125,7 +125,11 @@ func runLogin(cmd *cobra.Command, argv []string) (err error) {
 
 	if globalOpts.Manager {
 		logger.WithField("Cluster ID", clusterID).Debugln("Finding managing cluster")
-		clusterID, clusterName, err = ocm.DefaultOCMInterface.GetManagingCluster(clusterID)
+		var isHostedControlPlane bool
+		targetClusterID := clusterID
+		targetClusterName := clusterName
+
+		clusterID, clusterName, isHostedControlPlane, err = ocm.DefaultOCMInterface.GetManagingCluster(clusterID)
 		if err != nil {
 			return err
 		}
@@ -133,6 +137,17 @@ func runLogin(cmd *cobra.Command, argv []string) (err error) {
 		logger.WithFields(logger.Fields{
 			"ID":   clusterID,
 			"Name": clusterName}).Infoln("Management cluster")
+
+		// Print the related namespace if login to manager cluster
+		var namespaces []string
+		namespaces, err = listNamespaces(targetClusterID, targetClusterName, isHostedControlPlane)
+		if err != nil {
+			return err
+		}
+		fmt.Println("A list of associated namespaces for your given cluster:")
+		for _, ns := range namespaces {
+			fmt.Println("	" + ns)
+		}
 	}
 
 	if globalOpts.Service {
@@ -153,7 +168,7 @@ func runLogin(cmd *cobra.Command, argv []string) (err error) {
 			return fmt.Errorf("can't save the kube config into a specific location if multi-cluster is not enabled. Please specify --multi flag")
 		}
 		if _, err := os.Stat(args.kubeConfigPath); errors.Is(err, os.ErrNotExist) {
-			return fmt.Errorf("kube config save path is not exist")
+			return fmt.Errorf("the save path for the kubeconfig does not exist")
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


cmd/ocm-backplane/login/login.go
```diff
@@ -121,16 +121,56 @@ func runLogin(cmd *cobra.Command, argv []string) (err error) {
...
+
+	var env string
+	if strings.Contains(bpURL, "integration") {
+		env = "integration"
+	} else if strings.Contains(bpURL, "stage") {
+		env = "staging"
+	} else {
+		env = "production"
+	}
+
```
samanthajayasinghe, line 152: This may expose all available RH env's to the public. You may able to get the env with `ocm get /api/clusters_mgmt/v1/environment`  API or wait for [OSD-19188](https://issues.redhat.com/browse/OSD-19188) :)

cmd/ocm-backplane/login/login.go
```diff
@@ -352,3 +372,26 @@ func doLogin(api, clusterID, accessToken string) (string, error) {
...
+			hcpPrefix + clusterID + "-" + clusterName,
+		}
+	} else {
+		nsList = []string{
+			hivePrefix + clusterID,
+		}
+	}
+
+	return nsList
+}
```
samanthajayasinghe, line 397: What if we get the matching namespaces via bp-api endpoint ?  basically `oc get ns -A| grep {cluster-internal-id}`

pkg/utils/ocmWrapper.go
```diff
@@ -18,7 +18,7 @@ import (
 type OCMInterface interface {
 	IsClusterHibernating(clusterID string) (bool, error)
 	GetTargetCluster(clusterKey string) (clusterID, clusterName string, err error)
-	GetManagingCluster(clusterKey string) (clusterID, clusterName string, err error)
+	GetManagingCluster(clusterKey string) (clusterID, clusterName string, isHypershift bool, err error)
```
samanthajayasinghe, line 21: may be `isHostedControllerPlane`

feichashao: `isHostedControlPlane`

cmd/ocm-backplane/login/login.go
```diff
@@ -121,16 +121,56 @@ func runLogin(cmd *cobra.Command, argv []string) (err error) {
 		"ID":   clusterID,
 		"Name": clusterName}).Infoln("Target cluster")
 
+	// validate kubeconfig save path when login into multi clusters
+	if args.kubeConfigPath != "" {
+		if !args.multiCluster {
+			return fmt.Errorf("can't save the kube config into a specific location if multi-cluster is not enabled. Please specify --multi flag")
+		}
+		if _, err := os.Stat(args.kubeConfigPath); errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("kube config save path is not exist")
```
xiaoyu74, line 130: nit:
```suggestion
			return fmt.Errorf("the save path for the kube config does not exist")
```

cmd/ocm-backplane/login/login.go
```diff
@@ -121,16 +121,56 @@ func runLogin(cmd *cobra.Command, argv []string) (err error) {
 		"ID":   clusterID,
 		"Name": clusterName}).Infoln("Target cluster")
 
+	// validate kubeconfig save path when login into multi clusters
+	if args.kubeConfigPath != "" {
+		if !args.multiCluster {
```
xiaoyu74, line 126: Looks the current test doesn't cover the logic here based on the generated coverage info:
login_test_coverage

bmeng: I think that's true, but I did not touch this block... I just moved it to several lines above.

cmd/ocm-backplane/login/login.go
feichashao: The namespace pattern is fixed so it should be ok to print them directly.

cmd/ocm-backplane/login/login.go
bmeng: tried to implement part of this as well

cmd/ocm-backplane/login/login.go
samanthajayasinghe: Is this pattern unique to the FedRamp stack too?

bmeng:
/hold

Hold this for now, as we are waiting for an implementation which can get the env info automatically

PR #281, opened 2023-12-06
Fetch Backplane API URL via OCM environment 

samanthajayasinghe:
### What type of PR is this?

feature

This PR automatically detect the backplane URL via OCM env and users don't need set BP API URL in the config file 

https://issues.redhat.com/browse/OSD-19188
_Resolves #_

#### How to test 
**step 1**
* delete the URL setting from config file 
* Login prod/stg/int env clusters and check other bp-cli commands ( eg cloud console)

**step2** 
* Keep URL setting in the config file
* It will override via ocm env URL and warn users

**step3**
* Set BP API URL via env vars ( for ISV's )
* env BP API has high priority over OCM env

- [ * ] Ran unit tests locally
- [ * ] Validated the changes in a cluster


cmd/ocm-backplane/console/console_test.go
```diff
@@ -125,6 +126,7 @@ var _ = Describe("console command", func() {
 		dirName, _ := os.MkdirTemp("", ".kube")
 
 		pullSecretConfigDirectory = dirName
+		ocmEnv, _ = cmv1.NewEnvironment().BackplaneURL("https://backplane.api").Build()
```
supreeth7, line 129: Any reason why we're hardcoding the partial API string?

supreeth7: How about `https://dummy-api` :wink:

pkg/ocm/ocmWrapper.go
```diff
@@ -275,6 +276,23 @@ func (*DefaultOCMInterfaceImpl) GetStsSupportJumpRoleARN(ocmConnection *ocmsdk.C
 	return response.Body().RoleArn(), nil
 }
 
+// GetBackplaneURL returns Backplane URL based on ocm env
+func (*DefaultOCMInterfaceImpl) GetOCMEnvironment() (*cmv1.Environment, error) {
+	// Create the client for the OCM API:
+	connection, err := ocm.NewConnection().Build()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create OCM connection: %v", err)
```
supreeth7, line 284: Let's use the standard error format specifier for golang `%w`, applicable for error values, which wraps the error such that it can later be unwrapped with `errors.Unwrap`.

```suggestion
		return nil, fmt.Errorf("failed to create OCM connection: %w", err)
```

pkg/ocm/ocmWrapper.go
```diff
@@ -275,6 +276,23 @@ func (*DefaultOCMInterfaceImpl) GetStsSupportJumpRoleARN(ocmConnection *ocmsdk.C
 	return response.Body().RoleArn(), nil
 }
 
+// GetBackplaneURL returns Backplane URL based on ocm env
```
supreeth7, line 279: (nit)
```suggestion
// GetBackplaneURL returns the Backplane API URL based on the OCM env
```

pkg/ocm/ocmWrapper.go
```diff
@@ -275,6 +276,23 @@ func (*DefaultOCMInterfaceImpl) GetStsSupportJumpRoleARN(ocmConnection *ocmsdk.C
...
+func (*DefaultOCMInterfaceImpl) GetOCMEnvironment() (*cmv1.Environment, error) {
+	// Create the client for the OCM API:
```
supreeth7, line 281:
```suggestion
	// Create the client for the OCM API
```

pkg/cli/config/config.go
```diff
@@ -58,19 +60,30 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
...
 		return bpConfig, err
 	}
 
-	bpConfig.URL = viper.GetString("url")
+	// Check if user has explicitly defined backplane URL via env; it has higher precedence over the ocm env URL
+	url, isURLViaEnv := os.LookupEnv(info.BackplaneURLEnvName)
+	if isURLViaEnv {
+		logger.Infof("backplane URL set via env vars: %s", url)
+		bpConfig.URL = url
+	}
```
supreeth7, line 85: We don't need to explicitly do this, Viper does this for us automatically via `viper.AutomaticEnv()` :smiley: 

Ref: https://pkg.go.dev/github.com/spf13/viper#AutomaticEnv

pkg/cli/config/config.go
```diff
@@ -85,6 +98,21 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
...
+// GetBackplaneURL returns API URL
+func (config *BackplaneConfiguration) GetBackplaneURL() (string, error) {
+
+	ocmEnv, err := ocm.DefaultOCMInterface.GetOCMEnvironment()
+	if err != nil {
+		return "", err
+	}
+	url, ok := ocmEnv.GetBackplaneURL()
+	if !ok {
+		return "", fmt.Errorf("no backplane API defined for %v", ocmEnv.Name())
```
supreeth7, line 110: nit

```suggestion
		return "", fmt.Errorf("The requested API endpoint is not available for the OCM environment: %v", ocmEnv.Name())
```

pkg/cli/config/config.go
```diff
@@ -85,6 +98,21 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
...
+	}
+	logger.Infof("backplane URL fetch via OCM environment: %s", url)
```
supreeth7, line 112: Backplane is a noun, so let's abide by that :)
```suggestion
	logger.Infof("Backplane URL retrieved via OCM environment: %s", url)
```

pkg/cli/config/config.go
```diff
@@ -112,7 +140,11 @@ func (config BackplaneConfiguration) testHTTPRequestToBackplaneAPI() (bool, erro
 		http.DefaultTransport = &http.Transport{Proxy: http.ProxyURL(proxyURL)}
 	}
 
-	req, err := http.NewRequest("HEAD", config.URL, nil)
+	url, err := config.GetBackplaneURL()
+	if err != nil {
+		return false, err
+	}
```
supreeth7, line 146: nit: Let's make this boolean evaluation concise

```suggestion
	if url, err := config.GetBackplaneURL(); err != nil {
		return false, err
	}
```

pkg/cli/config/config.go
```diff
@@ -112,7 +140,11 @@ func (config BackplaneConfiguration) testHTTPRequestToBackplaneAPI() (bool, erro
...
+	req, err := http.NewRequest("HEAD", url, nil)
```
supreeth7, line 147: Let's use `http.Head()` instead, a refinement to our codebase.

```suggestion
	req, err := res, err := http.Head(url)
```

supreeth7: Also make it concise like the prev comment

supreeth7 (changes requested):
Overall nicely done! :100: 
Requested a few changes/refinements.

pkg/cli/config/config.go
samanthajayasinghe: if we use `viper.BindEnv("url", info.BackplaneURLEnvName)` then it will bind to the config file `url` key which may be deleted soon .. 
We will assign it to `bpConfig.URL` for easy usage.

pkg/cli/config/config.go
samanthajayasinghe: Will change it to `the` since 
```
pkg/cli/config/config.go:108:14: ST1005: error strings should not be capitalized (stylecheck)
                return "", fmt.Errorf("The requested API endpoint is not available for the OCM environment: %v", ocmEnv.Name())
```

pkg/cli/config/config.go
samanthajayasinghe: This change may cause to lint failure

```
pkg/cli/config/config.go:141:16: G107: Potential HTTP request made with variable url (gosec)
                if _, err := http.Head(url); err != nil {
```
https://stackoverflow.com/questions/70281883/golang-untaint-url-variable-to-fix-gosec-warning-g107

supreeth7: How about we skip the lint check by adding `//nolint`?

samanthajayasinghe: I think we can use `//nolint`  but we didn't use it before in bp-cli .. finally I didn't touch that function and kept it as it was .. 
but if you have a strong opinion we will change it with `//nolint`

pkg/cli/config/config.go
supreeth7: Ah, that didn't cross my mind, in that case, let's port the environment assignment into a modular function so that we can reuse it for other assignments in the future.

```
func getBackplaneEnv (key string) *select_return_type* {
        val, ok := os.LookupEnv(key)
        if ok {
            logger.Infof("backplane URL set via env vars: %s", url)
           return _
        }
    }
 ```
 The return type is on you to choose, you can either return the type `BackplaneConfiguration` or a `bool` to later evaluate it.

File: pkg/cli/config/config.go [hunk 10]
```diff
@@ -58,19 +60,29 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
...
 
-	// Check if user has explicitly defined proxy; it has higher precedence over the config file
-	err = viper.BindEnv("proxy-url", info.BackplaneProxyEnvName)
-	if err != nil {
+	// Warn user if url defined in the config file
+	if viper.GetString("url") != "" {
+		logger.Warn("URL configuration is deprecated, Pls remove backplane config URL setting")
+	}
+
+	// Fetch backplane URL from ocm env
```
supreeth7, line 74: Just noticed this, if I understand the flow correctly then the OCM API call is made regardless if the env var is set or not. How about we switch the flow?

i.e. **If** the env var is present then assign the config URL to that; **if else** then make the call to the OCM API endpoint. We'll be saving up on API calls in such case :smile:

supreeth7 (changes requested):

pkg/cli/config/config.go
samanthajayasinghe: That's a great point and fixed it ..

pkg/cli/config/config.go
samanthajayasinghe: Thanks for the generic function suggestion, and move into private func 
```
func getBackplaneEnv(key string) (string, bool) {
}
```

supreeth7 (approved):
/lgtm

PR #283, opened 2023-12-08
Use md5 hash as role session name when assuming into customer's support role

AlexVulaj:
### What type of PR is this?
feature

backplane-cli passes the SRE's email address as a role session name when assuming a customer's support role. We should not pass any SRE usernames to a customer when assuming into their account.

https://issues.redhat.com/browse/OSD-19901

- [x] Ran unit tests locally

### DIFF
--- a/cmd/ocm-backplane/cloud/common.go
+++ b/cmd/ocm-backplane/cloud/common.go
@@ -2,6 +2,7 @@ package cloud
 
 import (
 	"context"
+	//nolint:gosec
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -24,14 +25,17 @@ import (
 	logger "github.com/sirupsen/logrus"
 )
 
-const OldFlowSupportRole = "role/RH-Technical-Support-Access"
+const (
+	OldFlowSupportRole  = "role/RH-Technical-Support-Access"
+	CustomerRoleArnName = "Target-Role-Arn"
+)
 
 var StsClient = awsutil.StsClient
 var AssumeRoleWithJWT = awsutil.AssumeRoleWithJWT
 var NewStaticCredentialsProvider = credentials.NewStaticCredentialsProvider
 var AssumeRoleSequence = awsutil.AssumeRoleSequence
 
-// Wrapper for the configuration needed for cloud requests
+// QueryConfig Wrapper for the configuration needed for cloud requests
 type QueryConfig struct {
 	config.BackplaneConfiguration
 	OcmConnection *ocmsdk.Connection
@@ -207,7 +211,8 @@ func (cfg *QueryConfig) getCloudCredentialsFromBackplaneAPI(ocmToken string) (bp
 }
 
 type assumeChainResponse struct {
-	AssumptionSequence []namedRoleArn `json:"assumptionSequence"`
+	AssumptionSequence      []namedRoleArn `json:"assumptionSequence"`
+	CustomerRoleSessionName string         `json:"customerRoleSessionName"`
 }
 
 type namedRoleArn struct {
@@ -263,9 +268,15 @@ func (cfg *QueryConfig) getIsolatedCredentials(ocmToken string) (aws.Credentials
 		return aws.Credentials{}, fmt.Errorf("failed to unmarshal response: %w", err)
 	}
 
-	roleAssumeSequence := make([]string, 0, len(roleChainResponse.AssumptionSequence))
-	for _, namedRoleArn := range roleChainResponse.AssumptionSequence {
-		roleAssumeSequence = append(roleAssumeSequence, namedRoleArn.Arn)
+	assumeRoleArnSessionSequence := make([]awsutil.RoleArnSession, 0, len(roleChainResponse.AssumptionSequence))
+	for _, namedRoleArnEntry := range roleChainResponse.AssumptionSequence {
+		roleArnSession := awsutil.RoleArnSession{RoleArn: namedRoleArnEntry.Arn}
+		if namedRoleArnEntry.Name == CustomerRoleArnName {
+			roleArnSession.RoleSessionName = roleChainResponse.CustomerRoleSessionName
+		} else {
+			roleArnSession.RoleSessionName = email
+		}
+		assumeRoleArnSessionSequence = append(assumeRoleArnSessionSequence, roleArnSession)
 	}
 
 	seedClient := sts.NewFromConfig(aws.Config{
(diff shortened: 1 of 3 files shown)


pkg/awsutil/sts.go
```diff
@@ -25,8 +25,9 @@ const (
 	AwsConsoleURL                      = "https://console.aws.amazon.com/"
 	DefaultIssuer                      = "Red Hat SRE"
 
-	assumeRoleMaxRetries   = 3
-	assumeRoleRetryBackoff = 5 * time.Second
+	assumeRoleMaxRetries          = 3
+	assumeRoleRetryBackoff        = 5 * time.Second
+	assumeCustomerRoleSessionName = "RH-SRE"
```
samanthajayasinghe, line 30: Not a blocker, I feel that `bp-cli` may use ISV's for connecting to the cluster in the near future.  In that case, this logic may not be accurate. 
Ideally, we should move some of the assumed role logic to `bp-api`  endpoint where API decide how to determine the session-name based on some other factors.

cmd/ocm-backplane/cloud/common.go
```diff
@@ -24,14 +24,18 @@ import (
 	logger "github.com/sirupsen/logrus"
 )
 
-const OldFlowSupportRole = "role/RH-Technical-Support-Access"
+const (
+	OldFlowSupportRole         = "role/RH-Technical-Support-Access"
+	CustomerRoleArnName        = "Target-Role-Arn"
+	CustomerArnRoleSessionName = "RH-SRE"
```
samanthajayasinghe, line 30: This flow is already defined in `bp-api` 
https://gitlab.cee.redhat.com/service/backplane-api/-/blob/master/configs/cloud-config.yml?ref_type=heads#L9

AlexVulaj: Moving some of our slack conversation here for others who may look:

The assume role calls for the isolated flow occur here in the CLI. This is because SREs are using their local credentials which we don't want to send over the wire. There's a separate PR for backplane-api with the old flow

jharrington22: @AlexVulaj backplane today has a UUID used for service account creation on clusters, this UUID is mapped to an SRE. Are we able to use the same UUID?

AlexVulaj: @jharrington22  I'd be in favor of this, though I'm unfamiliar with the UUID you speak of. Do you know where this UUID is generated or how we do that mapping?

AlexVulaj: Never mind - found it in backplane API. I'm pushing a change now.

cmd/ocm-backplane/cloud/common.go
```diff
@@ -263,17 +267,24 @@ func (cfg *QueryConfig) getIsolatedCredentials(ocmToken string) (aws.Credentials
...
 
-	roleAssumeSequence := make([]string, 0, len(roleChainResponse.AssumptionSequence))
-	for _, namedRoleArn := range roleChainResponse.AssumptionSequence {
-		roleAssumeSequence = append(roleAssumeSequence, namedRoleArn.Arn)
+	assumeRoleArnSessionSequence := make([]awsutil.RoleArnSession, 0, len(roleChainResponse.AssumptionSequence))
+	for _, namedRoleArnEntry := range roleChainResponse.AssumptionSequence {
+		roleArnSession := awsutil.RoleArnSession{RoleArn: namedRoleArnEntry.Arn}
+		if namedRoleArnEntry.Name == CustomerRoleArnName {
+			data := []byte(email)
+			roleArnSession.RoleSessionName = fmt.Sprintf("%x", md5.Sum(data)) //nolint:gosec
```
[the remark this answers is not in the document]
AlexVulaj: It's definitely possible - it comes down to how quickly we want this out the door.

To go this route, we would add a new API endpoint that takes a string and returns an md5 hash, using the same existing functions that are used in the API today for performing that conversion. Then the bp cli would call that endpoint instead.

I'm happy to go that route, and then we can keep this PR in a "hold" until that API change gets released. Thoughts?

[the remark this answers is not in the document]
AlexVulaj: Actually another thought - since this is only for the new flow, we could update the response object of the existing API call to include a new property for the role session name to be used for the customer's support role. That would be substantially less overhead and allow us quicker turnover.

AlexVulaj:
/hold

See this comment: https://issues.redhat.com/browse/OSD-19901?focusedId=23636518&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-23636518

pkg/awsutil/sts.go
AlexVulaj: We've updated the API to determine the role session name that should be used, but the actual AssumeRole logic needs to be done client side since we use individual SRE credentials for the isolated backplane flow.

https://gitlab.cee.redhat.com/service/backplane-api/-/merge_requests/329

AlexVulaj:
/unhold

API changes have been promoted to production https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/92517

bmeng:
```
    "userIdentity": {
        "type": "AssumedRole",
        "principalId": "xxxxxx:RH-SRE-bmeng_sre.openshift",
        "arn": "arn:aws:sts::xxxxx:assumed-role/ManagedOpenShift-Support-98sn7d/RH-SRE-bmeng_sre.openshift",
        "accountId": "xxxxx",
        "accessKeyId": "xxxxx",
        "sessionContext": {
            "sessionIssuer": {
                "type": "Role",
                "principalId": "xxxxxx",
                "arn": "arn:aws:iam::xxxxx:role/ManagedOpenShift-Support-98sn7d",
                "accountId": "xxxxx",
                "userName": "ManagedOpenShift-Support-98sn7d"
            },
```
I still can see my user name in the cloudtrail, but it is not in the session part. I assume it is expected for now?

AlexVulaj:
@bmeng since the ticket specifically calls out the session name, I think we can go forward from here. I've started a thread here for further guidance on the principal being visible in cloudtrail:
https://redhat-internal.slack.com/archives/C058T1YEDPV/p1704395248986019

bmeng:
As discussed in slack, the scope for this pr should be limit the info from the session, for the rest parts we can cover them separately. 
/lgtm

PR #289, opened 2023-12-12
Allow to source --multi output

Tof1973:
### What type of PR is this?

feature

When using login --multi, the output cannot be sourced directly.
With this change we will be able to run:
 ```
source <(ocm-backplane login multi xxx)
```

_Resolves #_

- [ x] Ran unit tests locally
- [ x] Included documentation changes with PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,12 @@ Logging into multiple clusters via different terminal instances.
   $ export KUBECONFIG= <cluster-id-1-kube-config-path>
   ```
 
+  you can also directly run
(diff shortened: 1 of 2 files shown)


samanthajayasinghe:
Thanks for improving the usage of `multi` @Tof1973 ,
FYI , You can use `backplane session` for multi-cluster login where you don't need to set `kubecong` manually : [ref](https://github.com/openshift/backplane-cli#backplane-session)

PR #290, opened 2023-12-12
login to MC and SC when login to HCP

wanghaoran1988:
### What type of PR is this?

_(bug/feature/cleanup/documentation)_

This login to MC and SC when we login to HCP and add oc-mc oc-sc to access MC and SC

- [x] Ran unit tests locally

### DIFF
--- a/pkg/cli/session/session.go
+++ b/pkg/cli/session/session.go
@@ -59,54 +59,66 @@ var (
 
 // RunCommand setup session and allows to execute commands
 func (e *BackplaneSession) RunCommand(cmd *cobra.Command, args []string) error {
+	// Set Alias
 	if len(args) > 0 {
 		e.Options.Alias = args[0]
 	}
 	if e.Options.ClusterID == "" && e.Options.Alias == "" {
 		return fmt.Errorf("ClusterID or Alias required")
 	}
-
 	if e.Options.Alias == "" {
 		log.Println("No Alias set, using cluster ID")
 		e.Options.Alias = e.Options.ClusterID
 	}
-
-	// Verify validity of the ClusterID
 	clusterKey := e.Options.Alias
 	if e.Options.ClusterID != "" {
 		clusterKey = e.Options.ClusterID
 	}
-
 	clusterID, clusterName, err := ocm.DefaultOCMInterface.GetTargetCluster(clusterKey)
-
 	if err != nil {
-		return fmt.Errorf("invalid cluster Id %s", clusterKey)
-	}
-
-	clusterInfo, err := ocm.DefaultOCMInterface.GetClusterInfoByID(clusterID)
-	if err != nil {
-		return fmt.Errorf("failed to get cluster %s", clusterID)
+		return fmt.Errorf("failed to find cluster %s, %v", clusterKey, err)
 	}
 
 	// set cluster options
 	e.Options.ClusterName = clusterName
 	e.Options.ClusterID = clusterID
+
+	if e.Options.GlobalOpts.Manager && e.Options.GlobalOpts.Service {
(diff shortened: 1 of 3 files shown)


pkg/cli/session/session.go
```diff
@@ -312,6 +328,24 @@ set -euo pipefail
 	if err != nil {
 		return err
 	}
+	if e.Options.ManagementClusterID != "" {
+		// This add a command to access management cluster
+		err = e.createBin("oc-mc", "oc --kubeconfig="+filepath.Join(e.Path, e.Options.ManagementClusterID, "config")+" $@")
```
samanthajayasinghe, line 333: Shall we update session help text to show how to use `oc-mc` and `oc-sc` bins ? 

Ref : https://github.com/openshift/backplane-cli/blob/6411c2f1a7c07b07c538b15b76e423d2d8430760/pkg/cli/session/session.go#L417

mjlshen (commented):
Not really feedback on the code, but I don't want to login to the MC/SC when I `ocm backplane login` to an HCP. Is backplane-cli the right place for something like this?

E.g. should an SRE just build like a tmux script to run `ocm backplane session`, `ocm backplane session --manager`, and `ocm backplane session --service` if they want to do this?

wanghaoran1988:
> Not really feedback on the code, but I don't want to login to the MC/SC when I `ocm backplane login` to an HCP. Is backplane-cli the right place for something like this?
> 
`ocm backplane login` will not change, only the session command.
> E.g. should an SRE just build like a tmux script to run `ocm backplane session`, `ocm backplane session --manager`, and `ocm backplane session --service` if they want to do this?
Why a separate script? when you run "session", if it has all the related cluster logged in, why not?

mjlshen:
>>Not really feedback on the code, but I don't want to login to the MC/SC when I ocm backplane login to an HCP. Is backplane-cli the right place for something like this?

>ocm backplane login will not change, only the session command.

ack, good to know

>>E.g. should an SRE just build like a tmux script to run ocm backplane session, ocm backplane session --manager, and ocm backplane session --service if they want to do this?

>Why a separate script? when you run "session", if it has all the related cluster logged in, why not?

This is my personal opinion, but I think it's nice to have commands that do one thing well and let people compose them as they need/want to. For this case specifically, there's also somewhat of a security blast radius where an SRE is logging into three or two clusters at the same time; I think we should only get access to a cluster when we need to, not get access to all three clusters "just in case"

wanghaoran1988:
> > > Not really feedback on the code, but I don't want to login to the MC/SC when I ocm backplane login to an HCP. Is backplane-cli the right place for something like this?
> 
> > ocm backplane login will not change, only the session command.
> 
> ack, good to know
> 
> > > E.g. should an SRE just build like a tmux script to run ocm backplane session, ocm backplane session --manager, and ocm backplane session --service if they want to do this?
> 
> > Why a separate script? when you run "session", if it has all the related cluster logged in, why not?
> 
> This is my personal opinion, but I think it's nice to have commands that do one thing well and let people compose them as they need/want to. For this case specifically, there's also somewhat of a security blast radius where an SRE is logging into three or two clusters at the same time; I think we should only get access to a cluster when we need to, not get access to all three clusters "just in case"

Good point, but the login will do nothing except create a SA for user, and from what I can see now, when we login to HCP cluster, there is always a requirement to login to MC, sometime SC, and the HCP namespace is part of the HCP cluster.

bmeng:
@wanghaoran1988 is this still needed?

hectorakemp:
@wanghaoran1988 can we close this? Is this still relevant?

bmeng:
close this PR as it is stale for more than 5 months

PR #292, opened 2023-12-13
OSD-19779 Use local monitoring plugin on 4.14+ clusters

Tafhim:
### What type of PR is this?

bug

Since 4.14, the `monitoring-plugin` is required to be run locally. The plugin is running in cluster as a service so port forwarding is the only way to access it otherwise.
This PR sets up the locally running backplane console to include a locally running `monitoring-plugin` container in it's network and access it via a `nginx` server. The serer is pre-set in the `monitoring-plugin` container so we don't have other ways to access it.

This does not change the behavior for pre-4.14 clusters. It only adds the new container when the cluster is 4.14 and above.

[OSD-19779](https://issues.redhat.com//browse/OSD-19779)

This PR introduces a temporary nginx configuration and a function to run monitoring-plugin in a separate local container.
This change only takes effect on cluster with version 4.14 and above.

With this change:
- An additional container is run locally for monitoring-plugin
- The new container is run under the same network infrastructure of the console cotainer
- Both the console container and the monitoring-plugin container run in daemon mode
- Both containers use randomly reserved ports to listen for requests
- monitoring-plugin container uses an inbuilt nginx server that accepts a mounted nginx configuration. For this reason we generate a temporary nginx configuration in backplane configuration directory and mount that to the container.
- When both containers start up successfully, the terminal waits for an interrupt. Upon receiving the interrupt signal, both containers are stopped and removed.

- [x] Ran unit tests locally

### DIFF
--- a/cmd/ocm-backplane/console/console.go
+++ b/cmd/ocm-backplane/console/console.go
@@ -23,10 +23,12 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"os/signal"
 	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/Masterminds/semver"
@@ -44,6 +46,7 @@ import (
 	"k8s.io/client-go/rest"
 
 	"github.com/openshift/backplane-cli/pkg/cli/config"
+	"github.com/openshift/backplane-cli/pkg/info"
 	"github.com/openshift/backplane-cli/pkg/ocm"
 	"github.com/openshift/backplane-cli/pkg/utils"
 )
@@ -71,10 +74,17 @@ var (
 
 	// Pull Secret saving directory
 	pullSecretConfigDirectory string
+	containerNetwork          string
 )
 
 // Environment variable that indicates if open by browser is set as default
 const EnvBrowserDefault = "BACKPLANE_DEFAULT_OPEN_BROWSER"
+
+// Minimum required version for monitoring-plugin container
+const versionForMonitoringPlugin = "4.14"
+
+// Minimum required version to use backend service for plugins
+const versionForConsolePluginsBackendService = "4.12"
 
 // ConsoleCmd represents the console command
 var ConsoleCmd = &cobra.Command{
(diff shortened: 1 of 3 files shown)


cmd/ocm-backplane/console/console.go
```diff
@@ -214,6 +217,168 @@ func checkAndFindContainerURL(containerName string, containerEngine string) (err
...
+
+// setupMonitoringNginxConfig sets up a nginx configuration because
+// the monitoring container runs a nginx to accept requests
+func setupMonitoringNginxConfig(port string) error {
+	// setup the config file path
+	configDirectory, err := getBackplaneConfigDirctory()
+	if err != nil {
+		return err
+	}
+
```
samanthajayasinghe, line 238:
```suggestion
	if configDirectory, err := getBackplaneConfigDirctory(); err != nil {
		return err
	}

```

cmd/ocm-backplane/console/console.go
```diff
@@ -214,6 +217,168 @@ func checkAndFindContainerURL(containerName string, containerEngine string) (err
...
+events {}
+http {
+  include            /etc/nginx/mime.types;
+  default_type       application/octet-stream;
+  keepalive_timeout  65;
+  server {
+    listen              %s;
+    root                /usr/share/nginx/html;
+  }
+}
```
samanthajayasinghe, line 260: Any possibility of a YAML file?

cmd/ocm-backplane/console/console.go
```diff
@@ -214,6 +217,168 @@ func checkAndFindContainerURL(containerName string, containerEngine string) (err
...
+	`
+
+	config := fmt.Sprintf(configTemplate, port)
+
+	err = os.WriteFile(configFilename, []byte(config), 0600)
+	if err != nil {
+		return err
+	}
+
```
samanthajayasinghe, line 269:
```suggestion
	if err = os.WriteFile(configFilename, []byte(config), 0600); err != nil {
		return err
	}

```

cmd/ocm-backplane/console/console.go
```diff
@@ -214,6 +217,168 @@ func checkAndFindContainerURL(containerName string, containerEngine string) (err
...
+	// change permission as a work around to gosec
+	err = os.Chmod(configFilename, 0640)
+	if err != nil {
+		return err
+	}
+
```
samanthajayasinghe, line 275:
```suggestion
	if err = os.Chmod(configFilename, 0640); err != nil {
		return err
	}

```

cmd/ocm-backplane/console/console.go
```diff
@@ -675,6 +934,30 @@ func isRunningHigherThan411() bool {
...
+	if err != nil {
+		return false
+	}
+	clusterVersion := currentCluster.OpenshiftVersion()
+	if clusterVersion != "" {
+		version, err := semver.NewVersion(clusterVersion)
+		if err != nil {
+			return false
+		}
+		if version.Minor() >= 14 {
```
samanthajayasinghe, line 954: can you define  `4.13` version as constant and rename the function `isClusterVersionHigherThan( expectedVersion ) bool`

cmd/ocm-backplane/console/console.go
```diff
@@ -214,6 +217,168 @@ func checkAndFindContainerURL(containerName string, containerEngine string) (err
...
+		return err
+	}
+	configFilePath := filepath.Join(configDirectory, "monitoring-plugin-nginx.conf")
+
+	// Check if file already exists, if it does not, stop
+	if _, err = os.Stat(configFilePath); !os.IsNotExist(err) {
+		if err != nil {
+			return fmt.Errorf("failed to find the monitoring plugin nginx configuration: %s", err)
+		}
+	}
```
samanthajayasinghe, line 300: What if define the file name as constant in[ info.go](https://github.com/openshift/backplane-cli/blob/main/pkg/info/info.go) , 
is it good to move this logic as function into `pkg/cli/config`  package ?

samanthajayasinghe (commented):
This is a  great PR that fixes the console issue by installing a monitoring plugin. 
Can you pls add some unit tests to cover new methods?

cmd/ocm-backplane/console/console.go
```diff
@@ -417,6 +570,24 @@ func runConsole(cmd *cobra.Command, argv []string) (err error) {
 		documentationURL = "https://docs.openshift.com/rosa/"
 	}
 
+	if isRunningHigherThan413() {
+		monitoringContainerPortInt, err := utils.GetFreePort()
```
bmeng, line 574: This port is used by the console, and we won't access it directly right?

cmd/ocm-backplane/console/console.go
```diff
@@ -481,6 +652,48 @@ func runConsole(cmd *cobra.Command, argv []string) (err error) {
...
+			// forcing order of removal as the order is not deterministic between container engines
+			containersToCleanUp := []string{
+				monitoringPluginContainername,
+				consoleContainerName,
+			}
+
+			for _, c := range containersToCleanUp {
+				engStopArgs := []string{
+					"container",
+					"stop",
```
bmeng, line 672: Will this remove the container?

PR #295, opened 2023-12-15
Add Backplane URL deprecate env variable to control the warning message

samanthajayasinghe:
### What type of PR is this?
bug

backplane URL deprecate error message shows multiple time

_Resolves #_ https://issues.redhat.com/browse/OSD-20028

- [x] Ran unit tests locally

### DIFF
--- a/pkg/cli/config/config.go
+++ b/pkg/cli/config/config.go
@@ -68,7 +68,10 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
 
 	// Warn user if url defined in the config file
 	if viper.GetString("url") != "" {
-		logger.Warn("Manual URL configuration is deprecated, please remove URL key from Backplane configuration")
+		if _, ok := getBackplaneEnv(info.BackplaneURLDeprecateEnvName); !ok {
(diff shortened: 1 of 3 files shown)


supreeth7 (commented):
@samanthajayasinghe I can infer as `GetBackplaneConfiguration` is being called at two places in `pkg/utils/clientUtils` the warning is being displayed twice. 

This way of suppressing the warning message is quite a reach (aka hacky) and from a UX perspective, displaying the warning twice doesn't do any harm.

I propose we leave it as it is. The unit tests are good to stay though.

samanthajayasinghe:
> @samanthajayasinghe I can infer as `GetBackplaneConfiguration` is being called at two places in `pkg/utils/clientUtils` the warning is being displayed twice.
> 
> This way of suppressing the warning message is quite a reach (aka hacky) and from a UX perspective, displaying the warning twice doesn't do any harm.
> 
> I propose we leave it as it is. The unit tests are good to stay though.

Agreed with your point @supreeth7 , Anyway deprecation warning message is temporary for the next couple of versions.
I'll open a new PR only with tests then

PR #294, opened 2023-12-14
[OSD-20024] Allow GetRestConfigAsUser with elevationReason

Tof1973:
### What type of PR is this?

feature

### What this PR does / Why we need it?
When we use GetRestConfigAsUser (eg in osdctl) we cannot specift a reason for the elevation.

### Which Jira/Github issue(s) does this PR fix?

_Resolves #_

### Pre-checks (if applicable)

- [ x] Ran unit tests locally
- [ na] Validated the changes in a cluster
- [ na] Included documentation changes with PR

PR #298, opened 2023-12-18
[OSD-20024] Add test case for GetRestConfigAsUser

Tof1973:
### What type of PR is this?

_test_

followup of #294

_Resolves #_

- [ x] Ran unit tests locally
- [ na] Validated the changes in a cluster
- [ na] Included documentation changes with PR

### DIFF
--- a/cmd/ocm-backplane/login/login_test.go
+++ b/cmd/ocm-backplane/login/login_test.go
@@ -397,7 +397,7 @@ var _ = Describe("Login command", func() {
 
 	Context("check GetRestConfigAsUser", func() {
 
-		It("should fail if generated config do not contains provided username without elevationReasons",func () {
+		It("check config creation with username and without elevationReasons",func () {
 			mockOcmInterface.EXPECT().GetClusterInfoByID(testClusterID).Return(mockCluster, nil)
 			mockOcmInterface.EXPECT().GetOCMAccessToken().Return(&testToken, nil)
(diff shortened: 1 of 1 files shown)


Tof1973:
@samanthajayasinghe, here the PR to add test to cover changes in https://github.com/openshift/backplane-cli/pull/294#pullrequestreview-1782887545

cmd/ocm-backplane/login/login_test.go
```diff
@@ -387,4 +394,40 @@ var _ = Describe("Login command", func() {
 		})
 
 	})
+
+	Context("check GetRestConfigAsUser", func() {
+
+		It("should fail if generated config do not contains provided username without elevationReasons",func () {
```
typeid, line 400: The test name seems a bit confusing, I would reverse it: it should succeed if ... as you are also testing 
 `Expect(err).To(BeNil())` below. A test that mentions "should fail if..." should check for a failure.

cmd/ocm-backplane/login/login_test.go
```diff
@@ -387,4 +394,40 @@ var _ = Describe("Login command", func() {
...
+			username := "test-user"
+
+			config, err := GetRestConfigAsUser(backplaneConfiguration, testClusterID, username)
+			Expect(err).To(BeNil())
+			Expect(config.Impersonate.UserName).To(Equal(username))
+			Expect(len(config.Impersonate.Extra["reason"])).To(Equal(0))
+
+		})
+
+		It("should fail if generated config do not contains provided username/elevationReasons",func () {
```
typeid, line 415: Same as above.

PR #301, opened 2023-12-21
[OSD-20017] create config file with login --multi even if target config file exists

Tof1973:
### What type of PR is this?

bug

Without this change, when using login --multi on a cluster where we already did it, but where credentials expired, the connection will not work.

Resolves #[OSD-20017](https://issues.redhat.com//browse/OSD-20017)

- [ x] Ran unit tests locally
- [na ] Included documentation changes with PR

### DIFF
--- a/pkg/login/kubeConfig.go
+++ b/pkg/login/kubeConfig.go
@@ -35,7 +35,7 @@ func CreateClusterKubeConfig(clusterID string, kubeConfig api.Config) (string, e
 		}
 	}
 
-	// Write kube config if file not exist
+	// Write kube config
 	filename := filepath.Join(path, "config")
 	f, err := os.Create(filename)
 	if err != nil {
--- a/pkg/login/kubeConfig.go
+++ b/pkg/login/kubeConfig.go
@@ -41,18 +41,16 @@ func CreateClusterKubeConfig(clusterID string, kubeConfig api.Config) (string, e
 	if err != nil {
 		return "", err
 	}
-	err = clientcmd.WriteToFile(kubeConfig, f.Name())
+	defer func() {
+		f.Close()
+	}()
 
-	if err != nil {
-		return "", err
-	}
-	err = f.Close()
+	err = clientcmd.WriteToFile(kubeConfig, f.Name())
 	if err != nil {
 		return "", err
 	}
 
 	// set kube config env with temp kube config file
-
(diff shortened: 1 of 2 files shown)


pkg/login/kubeConfig.go
```diff
@@ -37,21 +37,18 @@ func CreateClusterKubeConfig(clusterID string, kubeConfig api.Config) (string, e
 
 	// Write kube config if file not exist
```
Dee-6777, line 38: A small change! Please change this comment as this PR is creating a new kubeconfig file even if it exits

Dee-6777 (changes requested):

pkg/login/kubeConfig.go
Tof1973: Oups, indeed ;)

pkg/login/kubeConfig.go
```diff
@@ -35,23 +35,20 @@ func CreateClusterKubeConfig(clusterID string, kubeConfig api.Config) (string, e
...
-			return "", err
-		}
-		err = f.Close()
-		if err != nil {
-			return "", err
-		}
+	if err != nil {
+		return "", err
+	}
+	err = f.Close()
```
typeid, line 49: **Suggestion:** use defer to call `f.Close()`. 
For the error handling: https://trstringer.com/golang-deferred-function-error-handling/

typeid (commented):
Generally lgtm, added one suggestion

pkg/login/kubeConfig.go
xiaoyu74: Yeah, I also thik using defer for calling `f.Close()` is better, it ensures that the file is closed as soon as the function `CreateClusterKubeConfig` returns, regardless of where the return happens.

xiaoyu74:
Also, given the changes made to the `CreateClusterKubeConfig` function that it now overwrites the kubeconfig file each time it is called, rather than only creating it if it doesn't exist, we might need to update test cases accordingly.

Apart from that, LGTM, great work @Tof1973 :+1:

pkg/login/kubeConfig.go
Tof1973: Done in latest commit...

bmeng:
Cannot understand why it has to check the file existence before writing... cc @samanthajayasinghe please let me know if there is a reason
/lgtm

PR #308, opened 2024-01-03
Always use isolated flow for HCP clusters

AlexVulaj:
### What type of PR is this?
feature

If we are dealing with an HCP cluster, there is no reason to call OCM to determine which flow to use. All HCP clusters must be on the new flow.

[OSD-20257](https://issues.redhat.com//browse/OSD-20257)

- [x] Ran unit tests locally

### DIFF
--- a/cmd/ocm-backplane/cloud/common.go
+++ b/cmd/ocm-backplane/cloud/common.go
@@ -56,7 +56,7 @@ func (cfg *QueryConfig) GetAWSV2Config() (aws.Config, error) {
 	return awsCreds.AWSV2Config()
 }
 
-// GetCloudCredentials returns Cloud Credentials Response
+// GetCloudConsole returns Cloud Credentials Response
 func (cfg *QueryConfig) GetCloudConsole() (*ConsoleResponse, error) {
 	ocmToken, _, err := cfg.OcmConnection.Tokens()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/ocm-backplane/cloud/common.go
```diff
@@ -281,6 +281,10 @@ func (cfg *QueryConfig) getIsolatedCredentials(ocmToken string) (aws.Credentials
 }
 
 func isIsolatedBackplaneAccess(cluster *cmv1.Cluster, ocmConnection *ocmsdk.Connection) (bool, error) {
+	if cluster.Hypershift().Enabled() {
```
mjlshen, line 284: The ticket description is:
> "I would like backplane-cli to always assume it should use the isolated backplane flow when dealing with an HCP cluster."

Why is the existing check insufficient? Also, is the statement that all HCP clusters use the isolated flow necessarily true, especially for older HCPs that were created before the isolated backplane flow was a thing?

AlexVulaj: The existing check _should_ be sufficient, but this saves an unnecessary call to OCM when we shouldn't need it. This in turn removes a failure point in the flow.

As for your second question, yes. It was a hard requirement of HCP GA to use this flow.

AlexVulaj: For more context: https://redhat-internal.slack.com/archives/CCX9DB894/p1704304764527919?thread_ts=1704181708.329699&cid=CCX9DB894

mjlshen: Could you double-check an ibm-infra HCP? I doubt these use the isolated backplane flow.

For that reason, I don't think we can save the OCM API call. If I'm wrong, then I'm wrong haha

AlexVulaj: Do you happen to have the cluster ID for that?

Also noteworthy - the code will still make a one-time fallback attempt to the old flow if the new flow fails.

AlexVulaj: I believe we can treat them as an edge case for now. The CLI will still fall back to make an attempt with the old flow if the new flow fails. Also, there's a thread going here about migrating IBM's managed policies: https://redhat-internal.slack.com/archives/CCX9DB894/p1704312385997819

fahlmant (approved):
/lgtm

PR #325, opened 2024-01-27
Detect assume-initial-arn from OCM URL

mjlshen:
### What type of PR is this?

feature

It can still be supplied in the config file as an override, but with this change it is now optional for individuals interacting with standard OCM URLs.

[OSD-20606](https://issues.redhat.com//browse/OSD-20606)

- [x] Ran unit tests locally

### DIFF
--- a/cmd/ocm-backplane/cloud/common.go
+++ b/cmd/ocm-backplane/cloud/common.go
@@ -221,6 +221,15 @@ type namedRoleArn struct {
 }
 
 func (cfg *QueryConfig) getIsolatedCredentials(ocmToken string) (aws.Credentials, error) {
+	const (
+		productionOCMUrl            = "https://api.openshift.com"
+		productionAssumeInitialArn  = "arn:aws:iam::922711891673:role/SRE-Support-Role"
+		stagingOCMUrl               = "https://api.stage.openshift.com"
+		stagingAssumeInitialArn     = "arn:aws:iam::277304166082:role/SRE-Support-Role"
+		integrationOCMUrl           = "https://api.integration.openshift.com"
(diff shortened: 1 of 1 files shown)


cmd/ocm-backplane/cloud/common.go
```diff
@@ -221,6 +221,15 @@ type namedRoleArn struct {
 }
 
 func (cfg *QueryConfig) getIsolatedCredentials(ocmToken string) (aws.Credentials, error) {
+	const (
+		productionOCMUrl            = "https://api.openshift.com"
+		productionAssumeInitialArn  = "arn:aws:iam::922711891673:role/SRE-Support-Role"
+		stagingOCMUrl               = "https://api.stage.openshift.com"
+		stagingAssumeInitialArn     = "arn:aws:iam::277304166082:role/SRE-Support-Role"
+		integrationOCMUrl           = "https://api.integration.openshift.com"
+		integrationAssumeInitialArn = "arn:aws:iam::277304166082:role/SRE-Support-Role"
```
supreeth7, line 230: I wouldn't recommend exposing internal API URLs in a public repository, that's the reason why we had to fallback to leverage CS environment resource for automation instead of doing it via the CLI, more details in https://issues.redhat.com/browse/OCM-1534.

mjlshen: Right, those are the backplane API URLs though, which I purposely avoided here

supreeth7: Thanks, I see it is safe to expose the ARNs publicly as well going by the Jira.

My initial assumption was they're something sensitive.

mjlshen: No worries, good to make sure since it looks very similar!

supreeth7 (approved):
/lgtm

PR #328, opened 2024-01-29
OSD-9023: Run managed scripts on hive/MC Cluster directly

kamlendu1982:
https://issues.redhat.com/browse/OSD-9023

create a cluster ExapleClusterName and login to it(ocm backplane login clusterid)
ocm backplane managedjob create --manager (This should result on running the job on the same hive as of the logged in cluster)
if --manager flag is set, it will fetch the hive id and replace ClusterID for the managed job with the hive/manager ID
All the hives are on production. So when we try this on staging, while retrieving the shard id, the OCM is still trying to find the hive in api.stage.openshift.com( this is the "url" variable in OCM config)
This is designed to work only on prod, so this feature will only be available on prod


cmd/ocm-backplane/managedJob/createManagedJob.go
```diff
@@ -63,6 +65,13 @@ func newCreateManagedJobCmd() *cobra.Command {
 		false,
 		"Fetch logs from the pod for the running job")
 
+	cmd.Flags().BoolVarP(
+		&options.manager,
+		"manager",
+		"",
+		false,
+		"Run the job on manager/hive shar if flag is set --manager")
```
bergmannf, line 73:
```suggestion
		"Run the job on manager/hive shard if flag is set --manager")
```

cmd/ocm-backplane/managedJob/createManagedJob.go
```diff
@@ -81,6 +90,19 @@ func runCreateManagedJob(cmd *cobra.Command, args []string) (err error) {
 	if err != nil {
 		return err
 	}
+	if options.manager {
+		if strings.Contains(bpCluster.ClusterURL, "api.stage") {
```
bergmannf, line 94: This is only relevant if the cluster is a normal OSD/ROSA cluster - for Hypershift clusters this should run just fine in staging.

bergmannf:
I tested this with a `hypershift` cluster in staging after a slight modification and it worked.
So looking good - I think you should be good to add the `manager` flag to the other backplane commands .

cmd/ocm-backplane/managedJob/createManagedJob.go
```diff
@@ -82,6 +90,26 @@ func runCreateManagedJob(cmd *cobra.Command, args []string) (err error) {
 		return err
 	}
 
+	if options.manager {
+		mcid, clusterName, _, err := ocm.DefaultOCMInterface.GetManagingCluster(bpCluster.ClusterID)
```
bergmannf, line 94:
```suggestion
		if mcid, clusterName, _, err := ocm.DefaultOCMInterface.GetManagingCluster(bpCluster.ClusterID); err == nil {
```

The first case can be handled in a single line.

cmd/ocm-backplane/managedJob/createManagedJob.go
```diff
@@ -82,6 +90,26 @@ func runCreateManagedJob(cmd *cobra.Command, args []string) (err error) {
...
+		} else {
+			c, err := ocm.DefaultOCMInterface.GetClusterInfoByID(bpCluster.ClusterID)
+			if err != nil {
+				return err
+			}
+			p, ok := c.GetProduct()
+			if !ok {
+				return fmt.Errorf("could not get product information")
+			}
+			return fmt.Errorf("product id is %s and bplane url is %s for cluster: %s\nThe feature is not available for OSD and ROSA", p.ID(), bpCluster.ClusterURL, clusterName)
```
bergmannf, line 109: I think this error is not entirely correct - as the not-supported combination would be an OSD/classic ROSA cluster, when using an environment that is not production (so either staging or integration).

Maybe the error could just indicate 'management cluster or shard not found in this OCM environment'.

bergmannf:
Didn't get around to retesting it yet, but I think this is fine now.

/lgtm

mjlshen (commented):

mjlshen (approved):
/lgtm

mjlshen (approved):
/approve

PR #329, opened 2024-02-01
OSD-19785. This adds multiple proxy support. 

tokalevasant:
This PR closed OSD-19785. Multiple proxies are saved as array in config file and the first working proxy is used.


samanthajayasinghe:
Thanks for the PR @tokalevasant . Can you pls add unit tests to cover the new function?

pkg/cli/config/config.go
```diff
@@ -96,6 +98,52 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
...
+func getFirstWorkingProxyURL(bpConfig BackplaneConfiguration, s []interface{}) string {
+	bpURL := bpConfig.URL + "/healthz"
+	logger.Infof("bpUrl: %s", bpURL)
+	for _, p := range s {
+		logger.Infof("proxy to check: %s", p)
+		if isProxyWorking(bpURL, p.(string)) {
+			return p.(string)
+		}
+	}
+	return ""
```
samanthajayasinghe, line 127: I guess you may need to return an error if any of the configured proxy is not functioning

pkg/cli/config/config.go
```diff
@@ -96,6 +98,52 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
 	return bpConfig, nil
 }
 
+func getProxyURL(bpConfig BackplaneConfiguration, proxyURLConfig any) string {
+	// as part of OSD-19785, proxy-url in config file is converted to []string from string to allow multiple proxies.
```
supreeth7, line 102: We do not need too descriptive comments, we can always traceback to the Jira lined to the commit if we need more details.

```suggestion
```

pkg/cli/config/config.go
```diff
@@ -96,6 +98,52 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
...
+	// For backward compatibility, get proxy-url as 'any' and check if it is still string (set in past) or array
+	// if string, return it else if array, return the first working proxy
+	switch t := proxyURLConfig.(type) {
+	case []interface{}:
+		logger.Infof("proxy-url array found in config file: %s ", t)
```
supreeth7, line 107: nit: `%s` -> `%v`

```suggestion
		logger.Infof("Multiple proxy URLs found in config file: %v ", t)
```

pkg/cli/config/config.go
```diff
@@ -96,6 +98,52 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
 	return bpConfig, nil
 }
 
+func getProxyURL(bpConfig BackplaneConfiguration, proxyURLConfig any) string {
```
samanthajayasinghe, line 101: you may not need to pass `bpConfig BackplaneConfiguration`  to return valid proxy.  what if implement the method as follows 
```
 func (config *BackplaneConfiguration) getValidProxy() {
    ... check if the proxy URL is a string or array here .....
 }
```

pkg/cli/config/config.go
```diff
@@ -96,6 +98,52 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
...
+	// as part of OSD-19785, proxy-url in config file is converted to []string from string to allow multiple proxies.
+	// For backward compatibility, get proxy-url as 'any' and check if it is still string (set in past) or array
+	// if string, return it else if array, return the first working proxy
+	switch t := proxyURLConfig.(type) {
+	case []interface{}:
+		logger.Infof("proxy-url array found in config file: %s ", t)
+		return getFirstWorkingProxyURL(bpConfig, proxyURLConfig.([]interface{}))
+	case interface{}:
+		logger.Warn("Change proxy-url in config file to an array. Multiple proxy-urls are now supported. First working proxy-url from the array will be used")
```
supreeth7, line 110: Are we trying to force the support for multiple proxy URLs for a user, this should ideally not be logged.

pkg/cli/config/config.go
```diff
@@ -96,6 +98,52 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
...
+		return proxyURLConfig.(string)
+	default:
+		logger.Warnf("Unrecognizable value type for proxy-url in config file: %s", t)
```
supreeth7, line 113: `Warnf` -> `Errorf`

pkg/cli/config/config.go
```diff
@@ -96,6 +98,52 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
...
+		return ""
+	}
+}
+
+func getFirstWorkingProxyURL(bpConfig BackplaneConfiguration, s []interface{}) string {
```
samanthajayasinghe, line 118: may be 
```
 func (config *BackplaneConfiguration) checkProxyHealth(string proxyURL) bool {
    ... check the proxy health for both cases ( array or string).....
 }
```

pkg/cli/config/config.go
```diff
@@ -96,6 +98,52 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
...
+		Transport: transport,
+		Timeout:   2 * time.Second,
+	}
+	req, _ := http.NewRequest("GET", bpURL, nil)
+	resp, err := client.Do(req)
+	if err != nil {
+		logger.Warnf("Proxy: %s gave error: %s", proxyURL, err)
+		return false
+	}
+	return resp.StatusCode >= 200 && resp.StatusCode < 300
```
samanthajayasinghe, line 144: you may able use http constants for `200` and `300` cases

pkg/cli/config/config.go
```diff
@@ -96,6 +98,52 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
...
+	proxy := http.ProxyURL(proxyURL)
+	transport := &http.Transport{Proxy: proxy}
+	client := &http.Client{
+		Transport: transport,
+		Timeout:   2 * time.Second,
+	}
+	req, _ := http.NewRequest("GET", bpURL, nil)
+	resp, err := client.Do(req)
+	if err != nil {
+		logger.Warnf("Proxy: %s gave error: %s", proxyURL, err)
```
supreeth7, line 141:
```suggestion
		logger.Errorf("Proxy: %s returned an error: %s", proxyURL, err)
```

pkg/cli/config/config.go
supreeth7, line 144: All successful HTTP connections should return only a `200` status code, any reason why we're checking for status codes under 300?

supreeth7 (changes requested):

pkg/cli/config/config.go
```diff
@@ -96,6 +98,52 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
...
+
+func isProxyWorking(bpURL, proxyToCheck string) bool {
+	proxyURL, _ := url.Parse(proxyToCheck)
+	proxy := http.ProxyURL(proxyURL)
+	transport := &http.Transport{Proxy: proxy}
+	client := &http.Client{
+		Transport: transport,
+		Timeout:   2 * time.Second,
+	}
+	req, _ := http.NewRequest("GET", bpURL, nil)
```
bergmannf, line 138: If we want to test if a proxy is working as an HTTP proxy, we might want to test `CONNECT` instead of `GET` as that's the method they *must* support.
However - not sure what happens if someone configures a `socks` proxy.

pkg/cli/config/config.go
tokalevasant: Fixed

pkg/cli/config/config.go
tokalevasant: Fixed. Returning error instead of logging

pkg/cli/config/config.go
tokalevasant: Fixed. As discussed on slack https://redhat-internal.slack.com/archives/C016S65RNG5/p1706887040612449

pkg/cli/config/config.go
tokalevasant: Fixed.

pkg/cli/config/config.go
tokalevasant: Fixed.

pkg/cli/config/config.go
tokalevasant: Fixed

pkg/cli/config/config.go
tokalevasant: The method needs to return first working proxy from the configured multiple proxies. I have changed code such that iif only one proxy is configured, it is converted to slice and passed to this method. Please elaborate if you meant something else.

pkg/cli/config/config.go
tokalevasant: Done

pkg/cli/config/config.go
tokalevasant: Done

tokalevasant:
> Thanks for the PR @tokalevasant . Can you pls add unit tests to cover the new function?

@samanthajayasinghe  I have added some tests and made all suggested changes. Can you please review changes again?

pkg/cli/config/config_test.go
```diff
@@ -59,3 +60,66 @@ func TestGetBackplaneConnection(t *testing.T) {
...
+		{"multiple-invalid-proxies", []interface{}{"-", "gellso", ""}, "", true},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			svr := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				_, _ = w.Write([]byte("dummy data"))
+			}))
+
+			t.Setenv("BACKPLANE_URL", svr.URL)
+			t.Setenv("HTTPS_PROXY", "http://dummy.proxy")
```
bergmannf, line 110: Is this set to test, that it's not defaulting to use the proxy environment variables?

bergmannf:
The changes look good to me, but it seems they are breaking an existing test:

```
 troubleshoot command
/go/src/github.com/openshift/backplane-cli/cmd/ocm-backplane/config/troubleshoot_test.go:17
  For backplane configuration
  /go/src/github.com/openshift/backplane-cli/cmd/ocm-backplane/config/troubleshoot_test.go:86
    should print the proxy url if backplane configuration is correct [It]
    /go/src/github.com/openshift/backplane-cli/cmd/ocm-backplane/config/troubleshoot_test.go:95
    Expected
              
        <int>: 1
    to equal
        <int>: 2
    /go/src/github.com/openshift/backplane-cli/cmd/ocm-backplane/config/troubleshoot_test.go:101 
```

pkg/cli/config/config.go
```diff
@@ -86,7 +87,18 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
 	bpConfig.AssumeInitialArn = viper.GetString("assume-initial-arn")
 
 	// proxyURL is optional
-	proxyURL := viper.GetString("proxy-url")
+	proxyInConfigFile := viper.Get("proxy-url")
```
AlexVulaj, line 90: If we use `viper.GetSlice` here, we can get a typed value back from the config instead of `any`. Then `getConfigProxyList` should be able to go away entirely.

pkg/cli/config/config.go
```diff
@@ -96,6 +108,52 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
...
+		logger.Infof("Multiple proxy URLs found in config file: %v ", t)
+		proxies = proxyURLConfig.([]interface{})
+	case interface{}:
+		proxies = []interface{}{proxyURLConfig}
+	}
+
+	return proxies, err
+}
+
+func (config *BackplaneConfiguration) getFirstWorkingProxyURL(s []interface{}) (string, error) {
```
AlexVulaj, line 126: I believe this function has broken some other tests - in a good way. Now that we are verifying that proxy URLs are legitimate URLs, we need to update our tests to make sure they aren't using things like "example-proxy"

pkg/cli/config/config_test.go
tokalevasant: Rewrote the test

tokalevasant:
> The changes look good to me, but it seems they are breaking an existing test:
> 
> ```
>  troubleshoot command
> /go/src/github.com/openshift/backplane-cli/cmd/ocm-backplane/config/troubleshoot_test.go:17
>   For backplane configuration
>   /go/src/github.com/openshift/backplane-cli/cmd/ocm-backplane/config/troubleshoot_test.go:86
>     should print the proxy url if backplane configuration is correct [It]
>     /go/src/github.com/openshift/backplane-cli/cmd/ocm-backplane/config/troubleshoot_test.go:95
>     Expected
>               
>         <int>: 1
>     to equal
>         <int>: 2
>     /go/src/github.com/openshift/backplane-cli/cmd/ocm-backplane/config/troubleshoot_test.go:101 
> ```

Fixed

pkg/cli/config/config.go
tokalevasant: I have refactored code to not throw error when no functioning proxy is found. This way existing functionality, of warning the user about proxy and moving on, is retained. All tests pass now. If we decide to throw error instead of warning, it breaks close to 20 tests.

pkg/cli/config/config.go
tokalevasant: Fixed. Thanks for this suggestion. Reduced code by a lot.

bmeng:
Hi @tokalevasant Can you try to squash the commits?

pkg/cli/config/config.go
```diff
@@ -96,6 +98,41 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
...
 }
 
+var clientDo = func(client *http.Client, req *http.Request) (*http.Response, error) {
+	return client.Do(req)
+}
+
+
+func (config *BackplaneConfiguration) getFirstWorkingProxyURL(s []string) (string) {
+	bpURL := config.URL + "/healthz"
+	logger.Infof("bpUrl: %s", bpURL)
```
supreeth7, line 108: I do not think we need to log this information. As it's logging redundant data.

```shell
INFO[0001] Backplane URL retrieved via OCM environment: https://<REDCATED> 
INFO[0001] bpUrl: https://<REDCATED>/healthz 
INFO[0003] Target cluster                                ID=<REDCATED>  Name=hs-mc-<REDCATED>
DEBU[0003] Using backplane URL: https://<REDCATED> 
```

supreeth7 (changes requested):

PR #330, opened 2024-02-02
Log relevant message when a user isn't able to fetch the API URL

supreeth7:
### What type of PR is this?
Bug

There are edge case scenarios where a user doesn't have enough permissions such as the BackplaneCLI OCM role to fetch the ClusterService Environment Resource from OCM CS.  This results in Backplane CLI abruptly timing out with a connection error message. This PR is an aim to address it from a UX perspective.

https://issues.redhat.com/browse/OSD-20675

- [ ] Ran unit tests locally

### DIFF
--- a/pkg/ocm/ocmWrapper.go
+++ b/pkg/ocm/ocmWrapper.go
@@ -2,6 +2,7 @@ package ocm
 
 import (
 	"fmt"
+	"net/http"
 	"strings"
 
 	"github.com/openshift-online/ocm-cli/pkg/ocm"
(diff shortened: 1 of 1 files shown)


pkg/ocm/ocmWrapper.go
```diff
@@ -289,6 +290,9 @@ func (*DefaultOCMInterfaceImpl) GetOCMEnvironment() (*cmv1.Environment, error) {
 
 	responseEnv, err := connection.ClustersMgmt().V1().Environment().Get().Send()
 	if err != nil {
```
xiaoyu74, line 292: Do we also need to consider an additional check for `responseEnv` being nil to handle cases where an error might prevent `responseEnv` from being populated as expected something like:? 
```
if err != nil {
    var isForbidden bool
    if responseEnv != nil {
        isForbidden = responseEnv.Status() == http.StatusForbidden
        if responseEnv.Error() != nil { 
            isForbidden = isForbidden || responseEnv.Error().Status() == http.StatusForbidden
        }
    }
    
    if isForbidden {
        return nil, fmt.Errorf("user does not have enough permissions to fetch the OCM environment resource. Please ensure you have the necessary permissions or try exporting the BACKPLANE_URL environment variable.\nUnderlying error: %w", err)
    }
    
    return nil, fmt.Errorf("failed to fetch the cluster environment: %w", err)
```

supreeth7: `responseEnv` will always be populated as going by the OCM Cluster Service codebase.

But, this is a nice suggestion, better to handle all the edge cases, let me add it.

mjlshen (approved):
/lgtm

PR #334, opened 2024-02-05
Console use tmp dir for monitor plugin conf

feichashao:
### What type of PR is this?

bug

Use the tmp directory instead of the config directory for monitoring plugin config.

The monitoring config doesn't contain sensitive info. The backplane config directory can sometimes be Read-only. So putting the file in tmp directory.

_Resolves #_

- [x] Ran unit tests locally

### DIFF
--- a/cmd/ocm-backplane/console/console.go
+++ b/cmd/ocm-backplane/console/console.go
@@ -59,7 +59,9 @@ type containerEngineInterface interface {
 	runMonitorPlugin(containerName string, consoleContainerName string, nginxConf string, pluginArgs []string) error
 }
 
-type podmanLinux struct{}
+type podmanLinux struct {
+	fileMountDir string
+}
 type podmanMac struct{}
 type dockerLinux struct{}
 type dockerMac struct{}
@@ -339,7 +341,7 @@ func (o *consoleOptions) getContainerEngineImpl() (containerEngineInterface, err
 
 	var containerEngineImpl containerEngineInterface
 	if runtime.GOOS == LINUX && containerEngine == PODMAN {
-		containerEngineImpl = &podmanLinux{}
+		containerEngineImpl = &podmanLinux{fileMountDir: filepath.Join(os.TempDir(), "backplane")}
 	} else if runtime.GOOS == MACOS && containerEngine == PODMAN {
 		containerEngineImpl = &podmanMac{}
 	} else if runtime.GOOS == LINUX && containerEngine == DOCKER {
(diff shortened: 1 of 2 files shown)


feichashao:
> @feichashao any possibilities to add some unit tests to cover the changes?

Hi @samanthajayasinghe , I was thinking the way to unit test those 2 functions but they are pretty OS dependent. I feel adding more function points for `os` may make the code less straightforward, and the test may just repeat the same logic of those 2 functions.   Any idea?

cmd/ocm-backplane/console/console.go
```diff
@@ -1172,29 +1169,26 @@ func (ce *dockerMac) runMonitorPlugin(containerName string, consoleContainerName
...
 // filename should be name only, not a path
 func (ce *podmanLinux) putFileToMount(filename string, content []byte) error {
-	// for files in linux, we put them into the user's backplane config directory
-	configDirectory, err := config.GetConfigDirctory()
-	if err != nil {
-		return err
-	}
-	dstFileName := filepath.Join(configDirectory, filename)
+	// for files in linux, we put them into the tmp directory
+	tmpDir := os.TempDir()
```
bmeng, line 1173: can we reuse the tmpDir in the other function?

bmeng: ok, seems it is not creating new dir in the tmp, then the question would be do we want to put the bp-cli related files into a separated directory under tempdir?

supreeth7 (commented):
Hey @feichashao, going through the review queue, is this PR stale?

cmd/ocm-backplane/console/console.go
samanthajayasinghe: Saving cluster specific nginx config into the `bp-cli-config` folder is not ideal. This will mainly block multi-cluster login since one console command writes cluster specific nginx conf to './config/backplane/nginx.conf' and other login override the same file. 

We should move files to `/tmp` or user defined folder and should rename the file with `{cluster-id}-nginx.config` that will reenable multi cluster login.

@feichashao regarding unit tests, if it's hard lets create a new card

cmd/ocm-backplane/console/console.go
```diff
@@ -1113,11 +1113,8 @@ func (ce *podmanMac) runMonitorPlugin(containerName string, consoleContainerName
 }
 
 func (ce *podmanLinux) runMonitorPlugin(containerName string, consoleContainerName string, nginxConf string, pluginArgs []string) error {
-	configDirectory, err := config.GetConfigDirctory()
-	if err != nil {
-		return err
-	}
-	nginxConfPath := filepath.Join(configDirectory, nginxConf)
+	tmpDir := os.TempDir()
+	nginxConfPath := filepath.Join(tmpDir, nginxConf)
```
samanthajayasinghe, line 1117: Shall we append {cluster-id} to nginx.conf, otherwise this file may override when the user connects to multi-cluster consoles

feichashao: That's a good point!  Let's put a cluster_id there.

cmd/ocm-backplane/console/console.go
bmeng: I mean we do not put it into the root of `/tmp`. But maybe something like `/tmp/backplane-cli/`, just to centralize the config files
If consider the cluster id as well, it may like `/tmp/backplane-cli/$CLUSTER_ID-nginx.conf`

cmd/ocm-backplane/console/console.go
```diff
@@ -59,7 +59,9 @@ type containerEngineInterface interface {
 	runMonitorPlugin(containerName string, consoleContainerName string, nginxConf string, pluginArgs []string) error
 }
 
-type podmanLinux struct{}
+type podmanLinux struct {
+	fileMountDir string
```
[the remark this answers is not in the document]
feichashao: The issue was reported for podman on linux so the aim of this PR is to provide a quick fix for it, I didn't test for other environments.
For Podman/Docker on Mac, the files are on a VM so the directory doesn't matter. For docker on linux I haven't tested. 
We can unify all the implementation to have `fileMountDir` though.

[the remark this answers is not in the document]
samanthajayasinghe: @feichashao, thanks for the reply. If it's not worth unifying the `fileMountDir` approach, we will keep it as it is.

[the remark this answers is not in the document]
feichashao: @samanthajayasinghe yeah, I think we can keep it as is. This is not a must for other environment.

PR #346, opened 2024-02-20
Do not override KUBECONFIG when RunElevate

Tof1973:
### What type of PR is this?
bug

When we call RunElevate, it overrides the definition of KUBECONFIG variable to a temporary file which is then deleted, which makes following kube commands to fail as there is no more configuration defined

_Resolves #[OSD-21061](https://issues.redhat.com//browse/OSD-21061)

Once this will be implemented and delivered, osdctl will be updated to remove hack (https://github.com/openshift/osdctl/pull/515)

- [ x] Ran unit tests locally
- [ na] Included documentation changes with PR

### DIFF
--- a/pkg/elevate/elevate.go
+++ b/pkg/elevate/elevate.go
@@ -55,6 +55,19 @@ func RunElevate(argv []string) error {
 		return err
 	}
 
+	// As WriteKubeconfigToFile(utils.CreateTempKubeConfig) is overriding KUBECONFIG,
+	// we need to store its definition in order to redefine it to its original (value or unset) when we do not need it anymore
+	oldKubeconfigPath, oldKubeconfigDefined := os.LookupEnv("KUBECONFIG")
+	defer func() {
+		if oldKubeconfigDefined {
+			logger.Debugln("Will set KUBECONFIG variable to original", oldKubeconfigPath)
+			os.Setenv("KUBECONFIG", oldKubeconfigPath)
+		} else {
+			logger.Debugln("Will unset KUBECONFIG variable")
+			os.Unsetenv("KUBECONFIG")
+		}
+	}()
+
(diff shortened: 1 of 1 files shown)


pkg/elevate/elevate.go
```diff
@@ -55,6 +55,19 @@ func RunElevate(argv []string) error {
 		return err
 	}
 
+	// As WriteKubeconfigToFile(utils.CreateTempKubeConfig) is overriding KUBECONFIG,
+	// we need to store its definition in order to redefine it to its original (value or unset) when we do not need it anymore
+	oldKubeconfigPath, oldKubeconfigDefined := os.LookupEnv("KUBECONFIG")
```
samanthajayasinghe, line 60: I guess, this may align with `login multi-cluster` subcommand and `session` commands

pkg/elevate/elevate.go
```diff
@@ -55,6 +55,19 @@ func RunElevate(argv []string) error {
...
+	defer func() {
+		if oldKubeconfigDefined {
+			logger.Debugln("Will set KUBECONFIG variable to original", oldKubeconfigPath)
+			os.Setenv("KUBECONFIG", oldKubeconfigPath)
```
samanthajayasinghe, line 64: any chance to add some unit tests to cover this change?
better to cover all login commands ( `login {cluster-id}` , `login --multi `  , `session {cluster-id}` )

pkg/elevate/elevate.go
Tof1973: No this is out of context of multi-cluster/session as it is only for elevate

pkg/elevate/elevate.go
Tof1973: Sorry, but this is not a login command, this will be used only for elevate command

PR #348, opened 2024-02-22
Add console tests to console.go 

rhs-harliu:
### What type of PR is this?
unit-test

Increase code coverage on console.go

https://issues.redhat.com/browse/OSD-20030

N/A 

- [ X] Ran unit tests locally


cmd/ocm-backplane/console/console.go
```diff
@@ -249,6 +247,11 @@ func (o *consoleOptions) run(cmd *cobra.Command, argv []string) error {
 	if err != nil {
 		return err
 	}
+	// If this is a test scenario, we throw an error as opposed of keeping this function in a infinite loop
+	if len(argv) != 0 && argv[0] == "console_GO_test" {
```
hectorakemp, line 251: Maybe use a constant here instead of the raw `console_GO_test` flag?

hectorakemp (changes requested):
A small change regarding using a constant instead of a raw string, but otherwise, it looks great. Well done @rhs-harliu!!

cmd/ocm-backplane/console/console.go
rhs-harliu: I'll make changes tomorrow

cmd/ocm-backplane/console/console.go
```diff
@@ -249,6 +247,11 @@ func (o *consoleOptions) run(cmd *cobra.Command, argv []string) error {
 	if err != nil {
 		return err
 	}
+	// If this is a test scenario, we throw an error as opposed of keeping this function in a infinite loop
```
bmeng, line 250: Why do we need to indicate the tests in the code here?

rhs-harliu: This is due to this function which gets called by cleanUp()


cmd/ocm-backplane/console.go
```
func execActionOnTermination(action postTerminationAction) error {
	sigs := make(chan os.Signal, 1)
	// syscall SIGUSR1 is used for testing
	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
	done := make(chan bool, 1)
	sig := <-sigs
	fmt.Println(sig)
	done <- true
	err := action()
	if err != nil {
		return err
	}
	return nil
}
```

Signal.Notify creates a busy wait which blocks the execution of the code. And if I send a CTRL+C via go-routines it will fail the test case. 

The indication of test stops the execution prior to calling clean up to prevent the tests to be stuck.

There are some possible alternative solutions, I thought about them but I don't think any of them are great so I choose the easiest which is what I did above:

1. Use SIGUSR1 as a signal to terminate the executable - Use of Signals is generally discouraged since we can't guarantee that SIGUSR1 is implemented on all platforms.
2. Use interfaces to create a mock function - Significant changes to the code required.

cmd/ocm-backplane/console/console.go
```diff
@@ -249,6 +247,11 @@ func (o *consoleOptions) run(cmd *cobra.Command, argv []string) error {
...
+	if len(argv) != 0 && argv[0] == "console_GO_test" {
+		return fmt.Errorf("test concluded sucessfully")
```
samanthajayasinghe, line 252: Basically, It's not recommended to add unit testing support code within the prod code. If prod code needs `console_GO_test` arg for unit testing means the actual code is not unit testable. 

May be something missing with function/component isolations.

cmd/ocm-backplane/console/console_test.go
```diff
@@ -108,9 +114,59 @@ var _ = Describe("console command", func() {
...
+		}
+		if err := os.Chmod("/tmp/tmp_bin/docker", 0777); err != nil {
+			fmt.Printf("Failed to chmod the file: %v\n", err)
+		}
+		return oldpath
+	}
+
+	removePath := (func(oldpath string) {
+		os.Setenv("PATH", oldpath)
+	})
```
samanthajayasinghe, line 166: Maybe move all test support functions to console_test_util.go file

cmd/ocm-backplane/console/console_test.go
```diff
@@ -294,4 +376,127 @@ var _ = Describe("console command", func() {
...
+			mockOcmInterface.EXPECT().GetOCMEnvironment().Return(ocmEnvironment, nil).AnyTimes()
+			o := consoleOptions{}
+			o.port = "1337"
+			o.url = "http://127.0.0.2:1447"
+			o.openBrowser = false
+			o.containerEngineFlag = DOCKER
+
+			argvstrs1 := []string{"console_GO_test"}
+			err := o.run(&cobra.Command{}, argvstrs1)
+			// There isn't any way to stop the o.run loop while testing, so we have to feed it a varible to force a termination
```
samanthajayasinghe, line 495: you may able to use `cancelContext` for the cmd execution 
pls refer : 
https://pkg.go.dev/github.com/spf13/cobra#Command.ExecuteContextC
https://blog.ksub.org/bytes/2019/10/07/using-context.context-with-cobra/

cmd/ocm-backplane/console/console_test.go
rhs-harliu: I'll make changes

cmd/ocm-backplane/console/console_test.go
rhs-harliu: I've implemented changes in a separate branch: 

https://github.com/rhs-harliu/backplane-cli/tree/harliu-console-tests-context

Alongside an alternative:
https://github.com/rhs-harliu/backplane-cli/tree/harliu-console-tests-interfaces

I'll discuss when I have time.

rhs-harliu:
Closing Pull Request, using a separate branch

PR #360, opened 2024-03-05
OSD-19615: Enhance backplane-cli verbose messages for login subcommand

rhs-harliu:
### What type of PR is this?

feature

This commit adds some more detailed debug printouts for the ocm-backplane login command, the printouts could be used by using  --verbosity=debug.

e.g 
```
./ocm-backplane --verbosity=debug login <cluster identification>
```

[OSD-19615](https://issues.redhat.com/browse/OSD-19615)

_Resolves #_

- X Ran unit tests locally
- X Validated the changes in a cluster

### DIFF
--- a/cmd/ocm-backplane/login/login.go
+++ b/cmd/ocm-backplane/login/login.go
@@ -77,30 +77,39 @@ func init() {
 
 func runLogin(cmd *cobra.Command, argv []string) (err error) {
 	var clusterKey string
-
+	logger.Debugf("Running Login Command ...")
+	logger.Debugf("Checking Backplane Version")
 	utils.CheckBackplaneVersion(cmd)
 
+	logger.Debugf("Extracting Backplane Cluster ID")
 	// Get The cluster ID
 	if len(argv) == 1 {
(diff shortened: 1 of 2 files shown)


pkg/utils/util.go
```diff
@@ -250,3 +250,11 @@ func CheckBackplaneVersion(cmd *cobra.Command) {
 
 	logger.WithField("Current version", info.Version).WithField("Latest version", latestVersion).Warn("Your Backplane CLI is not up to date. Please run the command 'ocm backplane upgrade' to upgrade to the latest version")
 }
+
+func PrettyPrint(i interface{}) string {
```
feichashao, line 254: This function is implemented but never referenced?  ( I like this function :D )

feichashao:
LGTM in general, thanks for making it more verbose for troubleshooting!   I left a comment regarding the `PrettyPrint` in utils.

pkg/utils/util.go
rhs-harliu: Oops, that was missed in clean up, it's a nifty function tho...

Removing.

PR #366, opened 2024-03-12
OSD-20400 Backplane-cli should validate config file after upgrade

rhs-harliu:
### What type of PR is this?

feature

Adds a warning if assume-initial-arn isn't defined in the backplane configuration file. The checks are done when the function getbackplaneconfig() is called in pkg/cli/config/config.go.

It does not terminate or produce an error since there are cases where these are not required for the operation of the backplane. 

The original jira ticket requested that the checks to be done after upgrades, but I think it's probably better to do it every time we read from the backplane config file, just in case something is fiddling with it during runtime. 

e.g 

```
$ ./ocm-backplane login cluster_name
WARN[0001] Your Backplane CLI is not up to date. Please run the command 'ocm backplane upgrade' to upgrade to the latest version  Current version= Latest version=0.1.25
WARN[0001] Manual URL configuration is deprecated, please remove URL key from Backplane configuration 
WARN[0003] assume-initial-arn is not defined in the backplane config file, if you recently updated or if your arn isn't recorded in OCM, please consider putting it into the backplane config file: /home/XXX/XXX/XXX/XXX.json 
$
```

_Resolves #_

[OSD-20400](https://issues.redhat.com/browse/OSD-20400)

- X Ran unit tests locally
- X Validated the changes in a cluster


pkg/cli/config/config.go
```diff
@@ -85,6 +85,10 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
 	bpConfig.SessionDirectory = viper.GetString("session-dir")
 	bpConfig.AssumeInitialArn = viper.GetString("assume-initial-arn")
 
+	if bpConfig.AssumeInitialArn == "" {
```
xiaoyu74, line 88: nit: please add test case to improve the coverage rate
```suggestion
	if bpConfig.AssumeInitialArn == "" {
```
image

xiaoyu74: Also, didn't we implement to detect `assume-initial-arn` from OCM URL? 
Looks it has been hard-coded: 
https://github.com/openshift/backplane-cli/blob/30def33dca39dc676d1c8675a7972f995c7afdb0/cmd/ocm-backplane/cloud/common.go#L223-L231

May I know why we still need to check in backplane's configuration for that arn?

samanthajayasinghe: As I understand, This is not the right way to get the assume-role ARN for the OCM env, It should get it via bp-api/ocm API endpoints, not hardcode in the bp-cli :( 

This is a public repo, and this exposes all bp-api URL's and role ARN's 

FYI : @bmeng / @supreeth7 / @cblecker

samanthajayasinghe:
@rhs-harliu , for the latest cli- version, only `proxy-url` is required all other configs are optional 
cc : @xiaoyu74

pkg/cli/config/config.go
xiaoyu74: I agree, exposing that OCM API endpoint could lead to unauthorized access attempts which could be risky. 

Also, exposing an ARN for an AWS IAM role is riskier since it's unique identifiers for AWS resources, and while an arn alone doesn't grant access without corresponding credentials, it can provide useful information to an unauthorized user which can still have potential risks.

rhs-harliu:
> @rhs-harliu , for the latest cli- version, only `proxy-url` is required all other configs are optional cc : @xiaoyu74

Ack, I'll make changes to select documentation which still has the arn thing as a mandatory field. It isn't on any public facing docs from my understanding. 

Looking back at the error from the original ticket:

```
Message: error creating console link unable to assume-role chain: could not assume support role in customer's account: AccessDenied: User: arn:aws:sts::710019948333:assumed-role/RH-Technical-Support-Access/RH-SRE-chcollin is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::918934589723:role/ManagedOpenShift-HCP-ROSA-Support-Role
        status code: 403, request id: 5ae9bf14-bf34-4531-b9ac-9a63dd15d0ac
```
I reckon the revised approach is adding a more detailed error message provided that this error was fired. Something like your aws arn lacks sufficient privileges, please either manually define your arn or make changes to the aws side?

PR #386, opened 2024-04-02
OSD-21419: Fixed console container cleanup issues

samanthajayasinghe:
### What type of PR is this?

bug

Fixing the console container cleanup issues for both docker and podman CRI

https://issues.redhat.com/browse/OSD-21419

- [x] Ran unit tests locally

### DIFF
--- a/cmd/ocm-backplane/console/console.go
+++ b/cmd/ocm-backplane/console/console.go
@@ -17,6 +17,7 @@ limitations under the License.
 package console
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"fmt"
@@ -57,6 +58,7 @@ type containerEngineInterface interface {
 	// some container engines have special handlings for different types of containers
 	runConsoleContainer(containerName string, port string, consoleArgs []string, envVars []envVar) error
 	runMonitorPlugin(containerName string, consoleContainerName string, nginxConf string, pluginArgs []string) error
+	containerIsExist(containerName string) (bool, error)
 }
 
 type podmanLinux struct{}
@@ -132,6 +134,8 @@ var (
 
 	// Pull Secret saving directory
 	pullSecretConfigDirectory string
+
+	ctx, cancel = context.WithCancel(context.Background())
 )
 
 func newConsoleOptions() *consoleOptions {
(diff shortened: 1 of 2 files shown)


cmd/ocm-backplane/console/console.go
```diff
@@ -1309,3 +1365,23 @@ func (ce *dockerLinux) stopContainer(containerName string) error {
 func (ce *dockerMac) stopContainer(containerName string) error {
 	return generalStopContainer(DOCKER, containerName)
 }
+
+// podman-exist for Linux
```
xiaoyu74, line 1369: Just my 2 cents. Looks the `containerIsExist` method is implemented separately for each struct (`podmanLinux`, `podmanMac`, `dockerLinux`, `dockerMac`), is that a bit redundancy? How about only write the logic for containerIsExist once and defining the container engine string `podman`, `docker` as a property of the structs themselves?
```
type baseContainerEngine struct {
    engineType string
}

func (b *baseContainerEngine) containerIsExist(containerName string) (bool, error) {
    return generalContainerIsExist(b.engineType, containerName)
}

type podmanLinux struct {
    baseContainerEngine
}
```
So, in case we were to add another container engine in the future, is this structure more clear?

cmd/ocm-backplane/console/console.go
```diff
@@ -1023,10 +1053,12 @@ func podmanRunConsoleContainer(containerName string, port string, consoleArgs []
 	}
 	engRunArgs = append(engRunArgs, consoleArgs...)
 	logger.WithField("Command", fmt.Sprintf("`%s %s`", PODMAN, strings.Join(engRunArgs, " "))).Infoln("Running container")
-	runCmd := createCommand(PODMAN, engRunArgs...)
+	//runCmd := createCommand(PODMAN, engRunArgs...)
```
xiaoyu74, line 1056: please remove the commented line here

xiaoyu74:
Great work Sam. I've built the binary with your commit and tested in an OSD `4.15.2` cluster. Following the reproduce steps and your PR can clean-up the console container properly.

cmd/ocm-backplane/console/console.go
```diff
@@ -637,14 +659,22 @@ func (o *consoleOptions) cleanUp(ce containerEngineInterface) error {
...
-			err := ce.stopContainer(c)
+			exist, err := ce.containerIsExist(c)
 			if err != nil {
-				return fmt.Errorf("failed to stop container %s: %v", c, err)
+				return fmt.Errorf("failed to check container exist %s: %v", c, err)
+			}
+			if exist {
+				err := ce.stopContainer(c)
+				if err != nil {
+					return fmt.Errorf("failed to stop container %s: %v", c, err)
```
xiaoyu74, line 672: how about add some more context for the return errors?
```
failed to stop container '%s' during cleanup process. Ensure the container engine is running and accessible
```

xiaoyu74:
As I can see the overall test coverage for `console.go` is 39.2%, as discussed before we probably need to do some coverage improvement in backplane maintenance EPIC some time?

cmd/ocm-backplane/console/console.go
samanthajayasinghe: Update error message, As I noted, `%v`  will throw the detailed exception as to why it failed

samanthajayasinghe:
> As I can see the overall test coverage for `console.go` is 39.2%, as discussed before we probably need to do some coverage improvement in backplane maintenance EPIC some time?

Yeah, We need to change the way CRI executing( create/stop/fetch) to fake it properly. I'll create new card use CRI with podman/docker go libs that may allow us to code proper unit tests

feichashao:
Works perfect in my Podman+MacOS environment.

1. It can remove 2 containers in a normal case.
```
INFO[0200] Container removed: monitoring-plugin-2adi24mc34eug59c1jqo8hvhelre7g9a 
INFO[0201] Container removed: console-2adi24mc34eug59c1jqo8hvhelre7g9a 
```

2. If I kill the monitor container manually, it can remove the left behind console container.

cmd/ocm-backplane/console/console.go
samanthajayasinghe: This is a good suggestion @xiaoyu74 , I feel that we need refactor most of the code to handle proper CRI( container runtime interface) and may belong to this PR . I have seen some more duplicate methods in console and will create a card for that.

PR #397, opened 2024-04-17
Increase the timeout to get working proxy url

xiaoyu74:
### What type of PR is this?

_(bug)_

https://redhat-internal.slack.com/archives/G01DC1VRAGK/p1713328523405079?thread_ts=1713327968.217359&cid=G01DC1VRAGK


feichashao:
Adding some context:
2 seconds timeout is not sufficient for a long delay proxy, leading the bp-cli fail to pick a proxy.

This PR serves as a quick fix to unblock the users.

PR #403, opened 2024-04-26
OSD-21347: Provide a link to the Dynatrace UI that the user can see the corresponding managedjob output directly.

tokalevasant:
### What type of PR is this?

feature

Provides a link to the Dynatrace UI that the user can see the corresponding managedjob output directly.

[OSD-21347](https://issues.redhat.com//browse/OSD-21347)
_Resolves #_

- [x] Ran unit tests locally


AlexVulaj:
I have tested this against a stage MC and it works as expected.

cmd/ocm-backplane/managedJob/createManagedJob.go
```diff
@@ -335,3 +352,53 @@ func getJobStatus(client BackplaneApi.ClientInterface, job *BackplaneApi.Job) (B
...
+		return
+	}
+	u := &unstructured.Unstructured{}
+	u.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "dynatrace.com",
+		Kind:    "DynaKube",
+		Version: "v1alpha1",
+	})
+
+	err = listDynaKube(cl,context.Background(), u, client.InNamespace("dynatrace"))
```
samanthajayasinghe, line 381: Managed scripts execute in all types of clusters ( `osd`, `MC`, `SC` , `HCP` etc) .. but `dynatrace` NS only available on MC clusters 
This may not be applicable for other type of clusters

samanthajayasinghe:
initiate some conversation about this PR requirement 
https://redhat-internal.slack.com/archives/C016S65RNG5/p1714155462938349

cmd/ocm-backplane/managedJob/createManagedJob.go
AlexVulaj: I tested this on an HCP as well and it still worked fine. If the function does not find a "dynatrace" namespace, it will just return an empty list, and skip over the `EachListItem` call on line 386.

Is there a better way to know if we're not on an MC so we can skip the call altogether?

xiaoyu74:
Please rebase and have all the CI checks passed first.

cmd/ocm-backplane/managedJob/createManagedJob.go
samanthajayasinghe: The `dynatarace` NS may be available on `osd`/ `rosa` clusters as well ( via CU deployment) , the best way to identify the MC cluster is through the labels. 
As I explained in the [Slack thread](https://redhat-internal.slack.com/archives/C016S65RNG5/p1714425548781719?thread_ts=1714155462.938349&cid=C016S65RNG5), this feature perfectly matches for the managed script rather than printing DT url while creating every managed job.

tokalevasant:
> Please rebase and have all the CI checks pass first. @tokalevasant

Done

cmd/ocm-backplane/managedJob/createManagedJob.go
xiaoyu74: Yeah, I agree it might be better to use cluster labels to check if the cluster is a MC **to avoid unnecessary API calls** for clusters where dynatrace is not applicable. 

So, we probably can update the code to include a check for the cluster type using labels before trying to list the dynatrace namespace?

bmeng: +1 to limit this option for MC only

AlexVulaj: I'm not opposed to the idea, but I was waiting for an answer to my above question. To share it again:
> Is there a better way to know if we're not on an MC so we can skip the call altogether?

AlexVulaj: I asked the OCM folks - it looks like we can use the endpoint `/api/clusters_mgmt/v1/clusters/<internal_cluster_id>/external_configuration/labels` and check for a label with a key of `ext-hypershift.openshift.io/cluster-type` and a value of `management-cluster`.

@tokalevasant we can work on making this update together.

cmd/ocm-backplane/managedJob/createManagedJob.go
```diff
@@ -335,3 +360,75 @@ func getJobStatus(client BackplaneApi.ClientInterface, job *BackplaneApi.Job) (B
...
+		return nil
+	})
+
+	if err != nil{
+		return "", err
+	}
+	return
+}
+
+func isManagedCluster() (bool, error){
```
AlexVulaj, line 414: Hey @tokalevasant , from our pairing session today, here's the new function we wrote that you can use in place of this:
```
func isManagedCluster() (bool, error) {
	//  and check for a label
	// with a key of ext-hypershift.openshift.io/cluster-type and a value of management-cluster
	connection, err := oocm.NewConnection().Build()
	if err != nil {
		return false, fmt.Errorf("failed to create OCM connection: %v", err)
	}
	defer connection.Close()

	labelsListResponse, err := connection.ClustersMgmt().V1().Clusters().Cluster(options.clusterID).ExternalConfiguration().Labels().List().Send()
	if err != nil {
		if labelsListResponse.Status() == 400 {
			return false, nil
		}
		return false, fmt.Errorf("failed to get label from ocm: %v", err)
	}

	for _, label := range labelsListResponse.Items().Slice() {
		if label.Key() == clusterTypeLabelKey {
			return label.Value() == clusterTypeLabelManagementCluster, nil
		}
	}

	return false, nil
}
```

Make sure you add the necessary constants as well to the top of the file:
```
const (
	clusterTypeLabelKey               = "ext-hypershift.openshift.io/cluster-type"
	clusterTypeLabelManagementCluster = "management-cluster"
)
```

tokalevasant: Changes completed, rebased and pushed

cmd/ocm-backplane/managedJob/createManagedJob.go
AlexVulaj: With the addition of the new function `isManagementCluster`, I believe we can resolve this thread.

xiaoyu74: @AlexVulaj - The updated functions' logic looks good to me.  I think we can unhold for now. BTW, have you tested to create a managed-scripts testjob against the MC to see if it works as expected?

bmeng:
@tokalevasant This PR need a rebase.

cmd/ocm-backplane/managedJob/createManagedJob.go
```diff
@@ -74,6 +89,31 @@ func newCreateManagedJobCmd() *cobra.Command {
...
+
+	labelsListResponse, err := connection.ClustersMgmt().V1().Clusters().Cluster(options.clusterID).ExternalConfiguration().Labels().List().Send()
+	if err != nil {
+		if labelsListResponse.Status() == 400 {
+			return false, nil
+		}
+		return false, fmt.Errorf("failed to get label from ocm: %v", err)
+	}
+
+	for _, label := range labelsListResponse.Items().Slice() {
```
bmeng, line 108: Do we need to check the length of the slice before iterating?

cmd/ocm-backplane/managedJob/createManagedJob.go
```diff
@@ -127,13 +167,27 @@ func runCreateManagedJob(cmd *cobra.Command, args []string) (err error) {
...
+	if err != nil{
+		logger.Warn("Could not determine if cluster is Managed Cluster")
+	}
+
+	if mc { // print dynatrace url for managed clusters only		
+		dtInstance, err := getDynatraceInstance() 
+		if err != nil {
+			logger.Info("Could not get Dynatrace instance")
+		} else {
+			defer fmt.Fprintf(cmd.OutOrStdout(), "\nCheck dynatrace logs at: %s\n", dtInstance)
```
bmeng, line 188: Why do we need a defer here?

cmd/ocm-backplane/managedJob/createManagedJob.go
```diff
@@ -335,3 +389,53 @@ func getJobStatus(client BackplaneApi.ClientInterface, job *BackplaneApi.Job) (B
...
+	if err != nil {
+		return
+	}
+	cl, err := client.New(cfg, client.Options{})
+
+	if err != nil {
+		return
+	}
+	u := &unstructured.Unstructured{}
+	u.SetGroupVersionKind(schema.GroupVersionKind{
```
bmeng, line 412: just to confirm, we do not need additional permission to list this?

cmd/ocm-backplane/managedJob/createManagedJob.go
AlexVulaj: @xiaoyu74 It was a bit ago, but yes I tested these changes and all looked good to me.

cmd/ocm-backplane/managedJob/createManagedJob.go
tokalevasant: @bmeng  The idea was to print this message as the last line on cli

cmd/ocm-backplane/managedJob/createManagedJob.go
tokalevasant: What is the  idiomatic way to do it in go? If the length of the slice is 0, there will be 0 iterations of the for loop so what is the additional need to check for len()?

cmd/ocm-backplane/managedJob/createManagedJob.go
tokalevasant: @bmeng We tested it with @AlexVulaj 's id and we did not need any additional permissions. @AlexVulaj  please confirm.

cmd/ocm-backplane/managedJob/createManagedJob.go
```diff
@@ -127,13 +167,27 @@ func runCreateManagedJob(cmd *cobra.Command, args []string) (err error) {
...
 	job, err := createJob(client)
-
+	
 	if err != nil {
 		return err
 	}
+	
+	mc, err := isManagementCluster()
+	if err != nil{
+		logger.Warn("Could not determine if cluster is Managed Cluster")
```
feichashao, line 180:
```suggestion
		logger.Warn("Could not determine if cluster is a Management Cluster")
```

feichashao: Small nit.  Managed Cluster != Management Cluster.

PR #404, opened 2024-04-27
OSD-20504: Add test case for config's get subcommand

TheUndeadKing:
### What type of PR is this?

unit-test

Adding  test case for config's get subcommand

### DIFF
--- a/cmd/ocm-backplane/config/get_test.go
+++ b/cmd/ocm-backplane/config/get_test.go
@@ -1,0 +1,50 @@
+package config
+
+import (
+	"fmt"
+
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/openshift/backplane-cli/pkg/ocm"
+	ocmMock "github.com/openshift/backplane-cli/pkg/ocm/mocks"
+	"github.com/spf13/cobra"
+)
+
+var _ = Describe("get command", func() {
+
+	var (
+		mockCtrl         *gomock.Controller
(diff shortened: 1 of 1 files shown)


TheUndeadKing:
- In Config folder, we've get.go, set.go & troubleshoot.go. We've test suite for troubleshoot.go, which cover 60%.
- The get & set func only return error.

```
// Get func
func getConfig(cmd *cobra.Command, args []string) error {

// Set func
func setConfig(cmd *cobra.Command, args []string) error { 
```

 I tried to write test case[load value for backplane url & doesn't load any value for  backplane url] for get function. In both test case the error return as nil.
```
// O/P:
url: https://backplane.example.com 
Return value:  <nil>
[..]
url:  
Return nil:  <nil>
```

cmd/ocm-backplane/config/get_test.go
```diff
@@ -0,0 +1,50 @@
...
+	var (
+		mockCtrl         *gomock.Controller
+		mockOcmInterface *ocmMock.MockOCMInterface
+	)
+
+	ocmEnv, _ := cmv1.NewEnvironment().BackplaneURL("https://backplane.example.com").Build()
+
+	ocmEnvNil, _ := cmv1.NewEnvironment().BackplaneURL("").Build()
+
+	cmd := &cobra.Command{Use: "get"}
```
Tessg22, line 26: The `cmd` variable could be initialized within the BeforeEach block to ensure that each test case starts with a fresh instance which could prevent potential side effects from previous tests.

TheUndeadKing: If i declare cmd inside before each block, it's not been exported as global variable. i.e, can't use it in the test case

xiaoyu74:
@TheUndeadKing - Are you still working on this PR?

samanthajayasinghe:
@TheUndeadKing pls address the PR comments or close the PR

PR #406, opened 2024-04-29
OSD-22639: prompt, store and reuse elevate reason

Tof1973:
### What type of PR is this?

_feature/documentation/unit-test_

[OSD-22639](https://issues.redhat.com/browse/OSD-22639)

When running ocm-backplate elevate with an empty string as reason, a prompt will be displayed (if possible) to request the reason.
When a reason is provided (as first argument, on prompt or from previous context), it will be stored in cluster context in order that if a reason is needed for elevation within the following 20 min for the same cluster, no prompt will be displayed and the stored context will be used.

_Resolves [OSD-22639](https://issues.redhat.com//browse/OSD-22639)_

After this change, we will be able to update SOPs using elevate to specify an empty reason, and command will prompt for the reason.

- [x] Ran unit tests locally

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ To setup the PS1(prompt) for bash/zsh, please follow [these instructions](https:
 | `ocm backplane console [flags]`                                             | Launch the OpenShift console of the current logged in cluster                            |
 | `ocm backplane cloud console`                                               | Launch the current logged in cluster's cloud provider console                            |
 | `ocm backplane cloud credentials [flags]`                                   | Retrieve a set of temporary cloud credentials for the cluster's cloud provider           |
-| `ocm backplane elevate <reason> -- <command>`                               | Elevate privileges to backplane-cluster-admin and add a reason to the api request        |
+| `ocm backplane elevate <reason> -- <command>`                               | Elevate privileges to backplane-cluster-admin and add a reason to the api request, this reason will be stored for 20min for future usage        |
 | `ocm backplane monitoring <prometheus/alertmanager/thanos/grafana> [flags]` | Launch the specified monitoring UI (Deprecated following v4.11 for cluster monitoring stack)                          |
 | `ocm backplane script describe <script> [flags]`                            | Describe the given backplane script                                                      |
 | `ocm backplane script list [flags]`                                         | List available backplane scripts |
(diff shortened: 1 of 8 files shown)


Tof1973:
[OSD-22250](https://issues.redhat.com/browse/OSD-22250)

iamkirkbater:
I'll give my +1 to this here - however it will be up to the backplane CLI maintainers to review further and merge.

/lgtm

samanthajayasinghe:
This is not a blocker, but what if include `Elevate` section in the readme and document what elevate command does, especially X mins timeouts

pkg/elevate/elevate_context.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/openshift/backplane-cli/pkg/utils"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
+)
+
+const (
+	elevateExtensionName             = "ElevateContext"
+	elevateExtensionRetentionMinutes = 20
```
samanthajayasinghe, line 18: This may be good candidate for backplane config , default to 20 mins and override from the user values

Tof1973: That could be a nice improvement we could make in the future, but which might require further discussions, as we do not want to have it too big.

Tof1973:
> This is not a blocker, but what if include `Elevate` section in the readme and document what elevate command does, especially X mins timeouts

I initially have not updated README? as there were no elevate section, but here I created a new dedicated elevate section with all details...

pkg/elevate/elevate_context.go
iamkirkbater: I'm personally -1 to making this user-configurable. If we set a time for this it should be consistent across SREs across the fleet.

I would, however, be open to the idea of having a "do not save elevation reason" configuration option that is individually configurable, so that you'd be prompted _every_ time and never persist the elevation reason that someone could opt-into.

I think of this like elevating to SUDO on a machine. The first time it will prompt for your password, and then there is a timeout where it will not prompt you. If you go a bit of time without using `sudo` it will re-prompt for your password again. The default timeout is configured for all SUDOERS on the machine via /etc/sudoers; however an individual user could `sudo -k` each time they run sudo and it would invalidate the timestamp and prompt them again.

Offhand I'm not aware of any use-case where this may make sense, but I can imagine someone has a workflow where saving the elevation reason is not in their own best interest and they'd like to personally turn it off.

Tof1973: Agree with you @iamkirkbater, I have added this idea to [OSD-22250](https://issues.redhat.com/browse/OSD-22250) comments for future improvements...
TY

README.md
```diff
@@ -268,6 +268,56 @@ Folowing command delete the session
...
+## Backplane elevate
+If you need to run some oc command(s) with elevation using backplane-cluster-admin user, you can use the elevate command for this.
+
+The elevate command requires a none empty reason for the elevation. When a reason is provided it will be used for future usage, in order you do not have to provide a reason for each elevation commands. The reasons are stored in the kubeconfig context, so it is valid only for the cluster for which it has been provided. When a reason is created/used, the last used reason timestamp is updated in the context, and the reason will be kept for 20min after its last usage, in order to avoid bad usage.
+
+When you use the elevate command with an empty reason, it will look if a non expired reason is stored in the current context for this server, and if there is one it will use it. If there is no reason stored in current context, then if the stdin and stderr are not redirected to pipe or file, a prompt will be done to ask for the reason.
+
+### Run an elevate command with reason
+```
+$ ocm-backplane evate 'OHSS-xxxxxx' -- get secret xxx
```
hectorakemp, line 280: Small typo here - should be `elevate`?

README.md
```diff
@@ -268,6 +268,56 @@ Folowing command delete the session
 ```
 ocm backplane session --delete <session-name>
 ```
+## Backplane elevate
+If you need to run some oc command(s) with elevation using backplane-cluster-admin user, you can use the elevate command for this.
+
+The elevate command requires a none empty reason for the elevation. When a reason is provided it will be used for future usage, in order you do not have to provide a reason for each elevation commands. The reasons are stored in the kubeconfig context, so it is valid only for the cluster for which it has been provided. When a reason is created/used, the last used reason timestamp is updated in the context, and the reason will be kept for 20min after its last usage, in order to avoid bad usage.
```
hectorakemp, line 274: non-empty reason?

README.md
```diff
@@ -268,6 +268,56 @@ Folowing command delete the session
...
+```
+$ ocm-backplane evate 'OHSS-xxxxxx' -- get secret xxx
+```
+The provided reason will be used for elevation, but also stored for future elevation on this cluster.
+If a reason was already stored in the current_context, then this provided reason will be added to it.
+
+### Run an elevate command with empty reason
+If you run the elevate command with an empty reason for the first time (or after the expiration), then you will be prompt for the reason if possible
+```
+$ ocm-backplane evate '' -- get secret xxx
```
hectorakemp, line 288: Same as above

hectorakemp (commented):
Great work @Tof1973!
All of this makes sense to me - but there's just a few nits in readme related to typos that are simple fixes. I also have tried testing it and can't quite get it to work the way it is expected to in the readme - see this slack thread: https://redhat-internal.slack.com/archives/C016S65RNG5/p1714745369106709?thread_ts=1714393862.376999&cid=C016S65RNG5

README.md
Tof1973: non-empty means a non-empty string ;)

README.md
Tof1973: '' is an empty string, so it is an empty reason

README.md
Tof1973: fix commited

README.md
Tof1973: Clarify this here: https://github.com/openshift/backplane-cli/blob/bb3749b074c1322e3393ac08ac81f7129b98e0e1/README.md?plain=1#L274

PR #408, opened 2024-05-02
Remove fallbacks to old flow for cloud access

AlexVulaj:
### What type of PR is this?
cleanup

Currently the bp-cli falls back to the old flow if it fails to assume cloud credentials/console access with the new flow. This was a good fallback initially, but at this point it masks legitimate errors and prevents SREs from knowing what's wrong. This PR removes that fallback in place of passing along logs.

[OSD-21438](https://issues.redhat.com//browse/OSD-21438)

- [x] Ran unit tests locally
- [ ] ~~Included documentation changes with PR~~

I tested these changes against non-STS clusters, as well as STS clusters using both the old and new flow.

### DIFF
--- a/cmd/ocm-backplane/cloud/common.go
+++ b/cmd/ocm-backplane/cloud/common.go
@@ -70,20 +70,14 @@ func (cfg *QueryConfig) GetCloudConsole() (*ConsoleResponse, error) {
 
 	isolatedBackplane, err := isIsolatedBackplaneAccess(cfg.Cluster, cfg.OcmConnection)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


xiaoyu74:
- Built a bp-cli binary based on this PR to run towards a previous cluster `osd-v4prod-aws` (old flow access method) we used to test and the old flow related functions can handle the access properly.

- Updated if-else block for the 2nd conditions in `IsolatedBackplaneAccess` looks clear:

```
	return supportRoleArn.Resource != OldFlowSupportRole, nil
	} else {
		return false, nil
	}
```

PR #413, opened 2024-05-12
OSD-22935: Remove use of dummy.com from test cases in backplane-cli

tokalevasant:
### What type of PR is this?
unit-test
_(bug/feature/cleanup/documentation/unit-test)_

Replaces "dummy.com" used in unit tests with "proxy.invalid"  because dummy.com is considered malicious
[OSD-22935](https://issues.redhat.com//browse/OSD-22935)
_Resolves #_

- [ ] Ran unit tests locally

### DIFF
--- a/pkg/cli/config/config_test.go
+++ b/pkg/cli/config/config_test.go
@@ -79,27 +79,27 @@ func TestBackplaneConfiguration_getFirstWorkingProxyURL(t *testing.T) {
 		},
 		{
 			name:    "valid-proxies",
-			proxies: []string{"https://dummy.com"},
+			proxies: []string{"https://proxy.invalid"},
 			clientDoFunc: func(client *http.Client, req *http.Request) (*http.Response, error) {
 				return &http.Response{StatusCode: http.StatusOK}, nil
 			},
-			want: "https://dummy.com",
+			want: "https://proxy.invalid",
 		},
 		{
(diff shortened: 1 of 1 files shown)


pkg/cli/config/config_test.go
```diff
@@ -79,27 +79,27 @@ func TestBackplaneConfiguration_getFirstWorkingProxyURL(t *testing.T) {
...
 			clientDoFunc: func(client *http.Client, req *http.Request) (*http.Response, error) {
 				return &http.Response{StatusCode: http.StatusOK}, nil
 			},
-			want: "https://dummy.com",
+			want: "https://proxy.invalid",
 		},
 		{
 			name:    "multiple-valid-proxies",
-			proxies: []string{"https://dummy.com", "https://dummy.proxy"},
+			proxies: []string{"https://proxy.invalid", "https://dummy.proxy.invalid"},
```
xiaoyu74, line 90: Looks the "https://dummy.proxy.invalid" still contains `dummy" keyword, should it be replaced as well?

AlexVulaj: Looking at the Jira card, I think we're okay here. The concern was that `dummy[.]com` is a malicious website that we don't want to be pinging, so as long we don't directly hit that URL we should be fine.

supreeth7 (approved):
/lgtm

PR #420, opened 2024-05-16
[OSD-21620] Add validation for the backplane config file

bmeng:
validation mandatory value proxy-url in backplane config validate the PD API KEY in config file if login via PD

### What type of PR is this?

_(/feature/)_

There are two parts for this PR:
1. Add validation to the backplane config
The original plan is to make the struct `BackplaneConfiguration` to have the mandatory fields as required, but it turns out the there are multiple ways to setup the proxy-url currently.
- ocm backplane login --proxy-url
- HTTPS_PROXY=xxxx ocm backplane login
- read through the config file

So it is not possible to change the struct. Adding a small function to make sure the proxy url was set by any of above methods, and exit early if it is not set.

2. Validate the pd api key when login via PD
Currently, the PD API Key can be configured in config file only, and the `--pd` will be used in login sub-command only. So, just add a simple condition to check the backplane config when `--pd` passed.

_Resolves #[OSD-21620](https://issues.redhat.com//browse/OSD-21620)_

- [ ] Ran unit tests locally

### DIFF
--- a/pkg/cli/config/config.go
+++ b/pkg/cli/config/config.go
@@ -15,9 +15,10 @@ import (
 	"github.com/openshift/backplane-cli/pkg/ocm"
 )
 
+// Please update the validateConfig function if there is any required config key added
 type BackplaneConfiguration struct {
 	URL              string  `json:"url"`
-	ProxyURL         *string `json:"proxy-url"` // Optional
+	ProxyURL         *string `json:"proxy-url"`
 	SessionDirectory string  `json:"session-dir"`
 	AssumeInitialArn string  `json:"assume-initial-arn"`
 	PagerDutyAPIKey  string  `json:"pd-key"`
@@ -61,6 +62,11 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
 		}
 	}
 
+	if err = validateConfig(); err != nil {
+		// FIXME: we should return instead of warning here, after the tests do not require external network access
+		logger.Warn(err)
+	}
+
 	// Check if user has explicitly defined proxy; it has higher precedence over the config file
 	err = viper.BindEnv("proxy-url", info.BackplaneProxyEnvName)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


bmeng:
The test passed locally, but got timeout in the CI...

pkg/cli/config/config.go
```diff
@@ -139,7 +144,19 @@ func (config *BackplaneConfiguration) getFirstWorkingProxyURL(s []string) string
 	return ""
 }
 
-func GetConfigDirctory() (string, error) {
+func validateConfig() error {
+	globalOpts := &globalflags.GlobalOptions{}
+
+	// Validate the proxy url
+	if viper.GetStringSlice("proxy-url") == nil && os.Getenv(info.BackplaneProxyEnvName) == "" && globalOpts.ProxyURL == "" {
+		return fmt.Errorf("proxy-url must be set explicitly in either config file or via the environment HTTPS_PROXY or the global option --proxy")
```
feichashao, line 152: I have concerns on making proxy mandatory. There are cases where we don't need/want a proxy. Eg, for testing environment or Fedramp environment.

bmeng: Agree with you that it should not be mandatory, anyway it is Warning for now https://github.com/openshift/backplane-cli/pull/420/files#diff-4e5b97588faff423da8ddd605ee2a4760f4f168d9bdb8ad220d4adc5823ef086R67-R68
The plan is to make the field required, but it seems for some use cases, we will still need the proxy unset.

feichashao: ah, ok, I missed the `logger.Warn(err)`.

pkg/cli/config/config_test.go
```diff
@@ -26,6 +26,19 @@ func TestGetBackplaneConfig(t *testing.T) {
 		}
 	})
 
+	t.Run("It should fail if the proxy-url is not configured", func(t *testing.T) {
```
feichashao, line 29: I expect the `GetBackplaneConfiguration()` should succeed if the proxy-url is not configured. I am also confused on why this test succeed.....

bmeng:
The globalFlag `--proxy` part is a bit tricky, it is used in the `login` subcommand only. And it has the validation within the login cmd. So I removed it from the config validating function.
And also updated the tests.

mjlshen (approved):
/lgtm

PR #421, opened 2024-05-17
OSD-23312: Refactor GetRestConfig() to allow passing an ocm connection programatically

typeid:
### What type of PR is this?
cleanup

CAD needs to use the GetRestConfig() function, but it doesn't allow passing an ocmConnection programatically. 
To not refactor the whole ocm package/usage, I dupliacted the functions to allow passing and initializing on the fly. 

- [ ] Ran unit tests locally

### DIFF
--- a/cmd/ocm-backplane/login/login.go
+++ b/cmd/ocm-backplane/login/login.go
@@ -20,6 +20,7 @@ import (
 
 	BackplaneApi "github.com/openshift/backplane-api/pkg/client"
 
+	ocmsdk "github.com/openshift-online/ocm-sdk-go"
 	"github.com/openshift/backplane-cli/pkg/backplaneapi"
 	"github.com/openshift/backplane-cli/pkg/cli/config"
 	"github.com/openshift/backplane-cli/pkg/cli/globalflags"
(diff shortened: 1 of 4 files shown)


cmd/ocm-backplane/login/login.go
mjlshen, line 1: The code changes look fine to me. One thing to consider if we also want to mirror the `*WithConn` naming for functions in this file like `GetRestConfigWithConn` to sidestep a breaking change for osdctl https://github.com/openshift/osdctl/blob/f475456b95544fce30378f4a7684903e0626efb1/cmd/common/helpers.go#L43-L47

It's ok not to do this, but just remember to submit a follow up PR eventually to osdctl to fix it there as well otherwise :D

typeid: Good point! I'll submit a followup when we release this in osdctl, I think it generally makes more sense to always pass the connection in the context of accessing the config programatically.

typeid: Ended up changing it because it's  too much duplicated code :)

mjlshen (approved):
/lgtm

PR #427, opened 2024-05-21
Add completion command for configuring CLI auto completion

AlexVulaj:
### What type of PR is this?
feature

Adds a "completion" command so that we can configure auto/tab completion with the backplane-cli


AlexVulaj:
/hold

This might be baked into Cobra naturally now

PR #450, opened 2024-06-10
OSD-23985: Refactor pagerduty client and add unit tests

samanthajayasinghe:
### What type of PR is this?

cleanup

Refactor pagerduty client and add unit tests

_Resolves #_

[### Special notes for your reviewer](https://issues.redhat.com/browse/OSD-23985)

- [x] Ran unit tests locally


pkg/pagerduty/pagerduty_test.go
```diff
@@ -0,0 +1,174 @@
...
+				Description: testClusterName,
+			}
+
+			mockPdClient.EXPECT().ListIncidentAlerts(testIncidentID).Return(alertResponse, nil).Times(1)
+			mockPdClient.EXPECT().GetServiceWithContext(context.TODO(), testServiceID, gomock.Any()).Return(serviceResponse, nil).Times(1)
+
+			alerts, err := pagerDuty.GetIncidentAlerts(testIncidentID)
+			Expect(err).To(BeNil())
+			Expect(len(alerts)).To(Equal(1))
+
```
Dee-6777, line 67: A small suggestion, shall we also add a small check using `formatAlert` function ? It's used inside a for loop in GetIncidentAlerts which is appending formatted alerts to the main alerts struct.
We can also add a check for `RateLimited`.

pkg/pagerduty/pagerduty_test.go
```diff
@@ -0,0 +1,174 @@
...
+		})
+
+		It("Should return cluster-id for a incident", func() {
```
Dee-6777, line 70: Should return cluster-id for an incident

pkg/pagerduty/pagerduty_test.go
```diff
@@ -0,0 +1,174 @@
...
+			}
+
+			mockPdClient.EXPECT().ListIncidentAlerts(testIncidentID).Return(alertResponse, nil).Times(1)
+
+			clusterID, err := pagerDuty.GetClusterID(testIncidentID)
+			Expect(err).NotTo(BeNil())
+			Expect(clusterID).To(Equal(""))
+		})
+
+		It("Should return first cluster-id for multiple incident", func() {
```
Dee-6777, line 112: Should return first cluster-id for multiple incidents

Dee-6777 (changes requested):
Great work!! Added a few comments and suggestions.

pkg/pagerduty/pagerduty.go
```diff
@@ -2,112 +2,154 @@ package pagerduty
...
 	}
 
-	clusterID, err := c.GetClusterIDFromAlertList(incidentAlerts)
+	clusterName := strings.Split(service.Description, " ")[0]
+
+	return clusterName, nil
+}
+
+// GetClusterID retrieves the cluster ID associated with the given incident ID.
+func (pd *PagerDuty) GetClusterID(incidentID string) (string, error) {
```
bmeng, line 140: should we change this to `GetClusterIDFromIncident` in case we may fetch the cluster from alert directly in the future

pkg/pagerduty/pagerduty.go
```diff
@@ -2,112 +2,154 @@ package pagerduty
...
+	incidentAlerts, err := pd.GetIncidentAlerts(incidentID)
 	if err != nil {
 		return "", err
 	}
 
-	return clusterID, nil
+	switch len(incidentAlerts) {
```
bmeng, line 146: will this cover the edge case which the alerts in the same incident are not from the same cluster?
we can return with error for this as it should not happen in general.

pkg/pagerduty/pagerduty.go
```diff
@@ -2,112 +2,154 @@ package pagerduty
...
+	formatAlert.Name = alert.Summary
+	formatAlert.Status = alert.Status
+	formatAlert.WebURL = alert.HTMLURL
 
-	detailsValue, ok := cefDetails["details"]
-	if !ok || detailsValue == nil {
-		return "", fmt.Errorf("missing or invalid 'details' field in Common Event Format Details")
-	}
+	// Check if the alert is of type 'Missing cluster'
+	isCHGM := alert.Body["details"].(map[string]interface{})["notes"]
```
bmeng, line 100: how can we identify it is CHGM with the `notes`?

pkg/pagerduty/pagerduty.go
```diff
@@ -2,112 +2,154 @@ package pagerduty
...
-	return clusterID, nil
+
+	return formatAlert, nil
 }
 
-// GetClusterID retrieves the cluster ID associated with the given incident ID.
-func (c *RealPagerDutyClient) GetClusterID(incidentID string) (string, error) {
-	incidentAlerts, err := c.PdClient.ListIncidentAlertsWithContext(context.TODO(), incidentID, pagerduty.ListIncidentAlertsOptions{})
+// GetClusterName interacts with the PD service endpoint and returns the cluster name string.
+func (pd *PagerDuty) GetClusterName(serviceID string) (string, error) {
```
bmeng, line 127: Why do we need the cluster name?

samanthajayasinghe: I may need cluster name when automating the elevate reason via cluster login

pkg/pagerduty/pagerduty.go
samanthajayasinghe: I added test case to cover that 
https://github.com/openshift/backplane-cli/pull/450/files#diff-957dde9e5fea16f1b4fda8327f293e8ffb26fb91d23fd369541d8af4ec320e0eR209-R219

pkg/pagerduty/pagerduty.go
samanthajayasinghe: Yeah .. return the error if all alerts not belongs to the same cluster and added test case to cover that

pkg/pagerduty/pagerduty.go
samanthajayasinghe: Done

pkg/pagerduty/pagerduty_test.go
samanthajayasinghe: Added separate test case to cover format func

pkg/pagerduty/pagerduty.go
```diff
@@ -2,112 +2,162 @@ package pagerduty
...
+	CreatedAt   time.Time
+	WebURL      string
+	ClusterID   string
+	ClusterName string
 }
 
-// RealPagerDutyClient implements the PagerDutyClient interface using the real PagerDuty client.
-type RealPagerDutyClient struct {
-	PdClient *pagerduty.Client
+const (
```
bmeng, line 26: Those consts are not used anywhere?

pkg/pagerduty/pagerduty.go
```diff
@@ -2,112 +2,162 @@ package pagerduty
...
+	// PagerDuty Incident Statuses
+	StatusTriggered    = "triggered"
+	StatusAcknowledged = "acknowledged"
+	StatusHigh         = "high"
```
bmeng, line 30: Maybe update the last two to UrgencyHigh/UrgencyLow, as they are not status.

pkg/pagerduty/pagerduty.go
samanthajayasinghe: Yeah .. I used some intests and removed and unused once

pkg/pagerduty/pagerduty.go
samanthajayasinghe: lets remove unused once

PR #452, opened 2024-06-12
[OSD-20658] Updated the ci-release script to ensure remote upstream repo is added

xiaoyu74:
### What type of PR is this?

_(bug)_

To ensure remote upstream repo is added based on the job failures logs I got: https://storage.googleapis.com/test-platform-results/logs/branch-ci-openshift-backplane-cli-main-combined-release/1800401130698575872/build-log.txt

https://issues.redhat.com/browse/OSD-20658

### DIFF
--- a/ci-release.sh
+++ b/ci-release.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
+# Manual release steps for reference: https://github.com/openshift/backplane-cli/blob/main/docs/release.md
+
 set -e
+
+REPO_URL="https://github.com/openshift/backplane-cli.git"
 
 # Extract version from VERSION.md
 VERSION=$(grep 'Version:' VERSION.md | awk '{print $2}')
@@ -19,6 +23,16 @@ fi
 # Git configurations
 git config user.name "CI release"
 git config user.email "ci-test@release.com"
+
+# Ensure the remote repository 'upstream' is set to the correct URL
+if git remote | grep -iq 'upstream'; then
+    current_url=$(git remote get-url upstream)
+    if [ "$current_url" != "$REPO_URL" ]; then
+        git remote set-url upstream $REPO_URL
+    fi
+else
+    git remote add upstream $REPO_URL
+fi
 
 # Ensure working on the latest main
 git fetch upstream


ci-release.sh
```diff
@@ -20,10 +22,29 @@ fi
 git config user.name "CI release"
 git config user.email "ci-test@release.com"
 
+# Ensure the remote repository 'upstream' is added
+if ! git remote | grep -q 'upstream'; then
```
bmeng, line 26: How was the remote set previously?
Do we need to confirm the upstream target is correct as well?

ci-release.sh
```diff
@@ -20,10 +22,29 @@ fi
...
 # Tagging the release
 git tag -a "v${VERSION}" -m "Release v${VERSION}"
 git push upstream "v${VERSION}"
+
+# Ensure GITHUB_TOKEN is set
+if [ -z "$GITHUB_TOKEN" ]; then
+    echo "Error: GITHUB_TOKEN is not set."
+    exit 1
+else
+    echo "GITHUB_TOKEN is set: $GITHUB_TOKEN"
```
bmeng, line 43: Better to not print the TOKEN

ci-release.sh
xiaoyu74: As we can see in the job [failure logs](https://storage.googleapis.com/test-platform-results/logs/branch-ci-openshift-backplane-cli-main-combined-release/1800401130698575872/build-log.txt): 
```
[36mINFO[0m[2024-06-11T05:34:23Z] Resolved source https://github.com/openshift/backplane-cli to main@6075bc81 
```
Looks the remote upstream was already resolved by the CI system but not pointing to a valid repo. I can update a bit to ensure repo is set to the correct url

ci-release.sh
```diff
@@ -20,10 +24,32 @@ fi
...
 git push upstream "v${VERSION}"
+
+# Ensure GITHUB_TOKEN is set
+if [ -z "$GITHUB_TOKEN" ]; then
+    echo "Error: GITHUB_TOKEN is not set."
+    exit 1
+fi
+
+# Export the GITHUB_TOKEN for use by goreleaser
+export GITHUB_TOKEN
```
bmeng, line 52: should it be `export GITHUB_TOKEN=xxxx`?

xiaoyu74: Actually we don't have to specify `GITHUB_TOKEN` and `make release` in the script since they are defined in the [CI configuration](https://github.com/xiaoyu74/release/blob/86ce2be6ae6ba40c4b4547cb8675d0d8cb513d21/ci-operator/config/openshift/backplane-cli/openshift-backplane-cli-main.yaml#L57-L61) already. 

I will update the script to remove those 2.

PR #463, opened 2024-06-26
OSD-24127: New accessrequest subcommand used to manage access requests used by the lockbox feature

Nikokolas3270:
### What type of PR is this?

_feature_

This PR allows SREP people to create access requests through the CLI.
Access requests are used on clusters on which the lockbox feature is feature is enabled.
Lockbox feature let the customer decide whether or not he/she want to grant us access to customer data.

Contributes to [OSD-24127](https://issues.redhat.com//browse/OSD-24127)

- [ ] Ran unit tests locally


pkg/utils/util.go
```diff
@@ -287,3 +292,27 @@ func AppendUniqNoneEmptyString(slice []string, element string) []string {
...
+
+	if jiratoken == "" {
+		return nil, fmt.Errorf("JIRA token is not defined")
+	}
+
+	tp := jira.PATAuthTransport{
+		Token: jiratoken,
+	}
+	return jira.NewClient(tp.Client(), JiraBaseURL)
+}
```
samanthajayasinghe, line 318: This function nicely matches to the backplane config module rather than the util func 
https://github.com/openshift/backplane-cli/blob/main/pkg/cli/config/config.go

But we should create a wrapper JIRA interface for unit test purposes 
something like pagerduty client 
https://github.com/openshift/backplane-cli/blob/main/pkg/pagerduty/client.go

pkg/utils/util.go
```diff
@@ -24,12 +24,17 @@ import (
...
+	"github.com/andygrunwald/go-jira"
+	"github.com/spf13/viper"
 )
 
 const (
 	ClustersPageSize             = 50
 	BackplaneAPIURLRegexp string = `(?mi)^https:\/\/api\.(.*)backplane\.(.*)`
 	ClusterIDRegexp       string = "/?backplane/cluster/([a-zA-Z0-9]+)/?"
+	JiraTokenConfigKey           = "jira_token"
+	JiraBaseURL                  = "https://issues.redhat.com"
```
samanthajayasinghe, line 37: IMO, We shouldn't hardcode the internal JIRA in public git repo 
You can get the base URL via jira client with a valid token

cmd/ocm-backplane/accessrequest/getAccessRequest.go
```diff
@@ -0,0 +1,109 @@
+package accessrequest
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	ocmcli "github.com/openshift-online/ocm-cli/pkg/ocm"
```
samanthajayasinghe, line 8: You may not need to import ocm-cli here , it's already available on ocm client 
https://github.com/openshift/backplane-cli/blob/4ce37a3a4a21e8b11f16d1ec69d9103239d0fe12/pkg/ocm/ocmWrapper.go#L9

cmd/ocm-backplane/accessrequest/getAccessRequest.go
```diff
@@ -0,0 +1,109 @@
...
+	return ocm.DefaultOCMInterface.GetClusterActiveAccessRequest(ocmConnection, clusterID)
+}
+
+func printAccessRequest(clusterID string, accessRequest *acctrspv1.AccessRequest) {
+	accessRequestStatus := accessRequest.Status()
+	accessRequestStatusState := acctrspv1.AccessRequestState("<Undefined>")
+
+	if accessRequestStatus != nil && accessRequestStatus.State() != "" {
+		accessRequestStatusState = accessRequestStatus.State()
+	}
```
samanthajayasinghe, line 63: IMO, You may able to move these logics to pkg/access-request module and write some unit tests. 
otherwise it may be hard to write tests with cobra dependencies

cmd/ocm-backplane/accessrequest/createAccessRequest.go
```diff
@@ -0,0 +1,253 @@
...
+		return fmt.Errorf("failed to create OCM connection: %v", err)
+	}
+	defer ocmConnection.Close()
+
+	accessRequest, err := getAccessRequest(clusterID, ocmConnection)
+
+	if err != nil {
+		return err
+	}
+
```
samanthajayasinghe, line 87: You may not need to build ocm connection here, it may be part of ocm interface implementation 
ref : https://github.com/openshift/backplane-cli/blob/4ce37a3a4a21e8b11f16d1ec69d9103239d0fe12/pkg/ocm/ocmWrapper.go#L61

cmd/ocm-backplane/accessrequest/createAccessRequest.go
```diff
@@ -0,0 +1,253 @@
...
+				logger.Warnf("won't verify the validity of the '%s' JIRA issue as there is no issue service in the client", options.notificationIssueID)
+			} else {
+				notificationIssue, _, err = issueService.Get(notificationIssueID, nil)
+				if err != nil || notificationIssue == nil {
+					return fmt.Errorf("the '%s' JIRA issue is invalid: %v", options.notificationIssueID, err)
+				}
+			}
+		}
+	} else {
+		jiraClient, err := utils.GetJiraClient()
```
samanthajayasinghe, line 164: IMO , You may not be able to write unit test if you init JIRA client when executing the command 
It's better to init the JIRA client with an interface, which may help mock the JIRA downstream and test it.

samanthajayasinghe (commented):
Thanks, @Nikokolas3270 for the PR ... The concepts are really good 
I have some concerns with hard-coded JIRA/API URLs in the PR.

IMO, It's good to init JIRA client as an interface that may help to write unit tests.

Please try to cover at least 50% new code with unit tests

cmd/ocm-backplane/accessrequest/createAccessRequest.go
```diff
@@ -0,0 +1,253 @@
...
+	ocmcli "github.com/openshift-online/ocm-cli/pkg/ocm"
+	"github.com/openshift/backplane-cli/pkg/elevate"
+	"github.com/openshift/backplane-cli/pkg/ocm"
+	"github.com/openshift/backplane-cli/pkg/utils"
+	logger "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+const (
+	ProductionOCMUrl = "https://api.openshift.com"
```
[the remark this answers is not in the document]
Nikokolas3270: Thanks for the tip. I think I will actually call this method:
https://github.com/openshift/backplane-cli/blob/main/pkg/ocm/ocmWrapper.go#L297
I am indeed only interested to know whether we are in prod or not.

Note that I copy pasted this const def from there:
https://github.com/openshift/backplane-cli/blob/4ce37a3a4a21e8b11f16d1ec69d9103239d0fe12/cmd/ocm-backplane/cloud/common.go#L214
In other words: there are other existing places in the code where we might want to get rid of this hard code.

pkg/utils/util.go
Nikokolas3270: Sure I can move it there. However this `GetJiraClient` won't use any of the internal members of `BackplaneConfiguration`. So it won't be an object method there.

Ok to add some interface layer for the sake of mocking.
Will check if one already exists in the JIRA module.

pkg/utils/util.go
Nikokolas3270: The `JiraTokenConfigKey` is used with `viper` to retrieve the token from the config. So this one will remain a constant there.

For `JiraBaseURL`, this is used in case the CLI user calls option `--notification-issue` with a URL (ex: `https://issues.redhat.com/browse/SDAINT-9165`) instead of a JIRA issue ID (`SDAINT-9165`). I will see if maybe some proper regexp matching could be used to get the issue ID; if not: yes I could indeed use the JIRA client to get that URL prefix info; thanks for the tip.

cmd/ocm-backplane/accessrequest/getAccessRequest.go
Nikokolas3270: Not sure to understand you.
This package is imported on the following files:
```
> grep -rFl github.com/openshift-online/ocm-cli/pkg/ocm . | grep -vF _test | grep -vFi mock
./cmd/ocm-backplane/accessrequest/getAccessRequest.go
./cmd/ocm-backplane/accessrequest/createAccessRequest.go
./cmd/ocm-backplane/cloud/credentials.go
./cmd/ocm-backplane/cloud/console.go
./ocm-backplane
./pkg/ocm/ocmWrapper.go
```
File in imported in the commands which wish to perform several OCM call in the row with the same OCM client because, I guess, it is taking time to re-create the client.

In the new `ocm backplane accessrequest create` command, this is exactly what we are doing:
1. We first check the existence of an existing access request
2. If there is none, we create one.

Both OCM calls will use the same client. The first call will be done through the `accessrequest.getAccessRequest` method which needs the OCM client as a parameter.

For the sake of factorisation, the same method is called in the `ocm backplane accessrequest get` handled by this file.  
Hence the OCM client needs to be created prior calling that method by calling `ocmcli.NewConnection().Build()`, hence the need to import that package.

cmd/ocm-backplane/accessrequest/createAccessRequest.go
Nikokolas3270: As said in the "import `github.com/openshift-online/ocm-cli/pkg/ocm` package" comment:
- The `create` subcommand may need to perform several calls in the row to OCM, hence it is probably best to not re-create the OCM client between the calls. Right?
- Hence the OCM wrapper methods interacting with the access transparency sub-API of OCM need to take the OCM client as parameter.
- Last but not least: some other methods of the OCM wrapper already take the OCM client as parameter, probably for the same reason: repetitive calls to the OCM API.

cmd/ocm-backplane/accessrequest/getAccessRequest.go
Nikokolas3270: Agreed, I may need to move the code a bit for unittest purposes.

cmd/ocm-backplane/accessrequest/createAccessRequest.go
Nikokolas3270: Yes, for unittests mocking, the JIRA client will need to implement some interface.

cmd/ocm-backplane/accessrequest/createAccessRequest.go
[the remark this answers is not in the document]
Nikokolas3270: Actually, digging further, the proposed method `bpconfig.GetBackplaneURL()` wouldn't have helped to determine if we are in prod or not.
Indeed the method just retrieve the URL to OCM without saying if it is prod or not.
Sticking to calling `GetOCMEnvironment` and comparing the returned env name to `production` for that test.

pkg/utils/util.go
Nikokolas3270: Also, concerning `JiraBaseURL`, the constant is used to initialise the JIRA client... so we definitely cannot get rid of it I guess.

cmd/ocm-backplane/accessrequest/createAccessRequest.go
```diff
@@ -0,0 +1,253 @@
...
+		}
+
+		issueType := "Story"
+		if jiraProject == "OHSS" {
+			issueType = "Incident"
+		}
+
+		notificationIssue = &jira.Issue{
+			Fields: &jira.IssueFields{
+				Description: "To be defined",
```
wanghaoran1988, line 183: Let's just put "Access request tracker" ?

wanghaoran1988: Could we also add some labels so we can do filter easily?

pkg/utils/util.go
samanthajayasinghe: You are right, @Nikokolas3270 , We may need to specify `JiraBaseURL` 
As a hack, we can move the Jira base URL to the backplane config as well. 
You can decide what's best in this case :)

cmd/ocm-backplane/accessrequest/getAccessRequest.go
samanthajayasinghe: My thinking was to encapsulate the `ocm.NewConnection().Build()` into the `ocmWrapper` interface, where all ocm connections and implementation go to one module. 
If there are multiple call issues, we can improve the `ocmWrapper` interface implementation by caching the connection. 
anyway, you can decide what is the best approach for this scenario

cmd/ocm-backplane/accessrequest/createAccessRequest.go
samanthajayasinghe: Yeah, I understand your points about the OCM wrapper. We could probably refactor the OCM wrapper to cache the connection and reuse it in subsequent calls. I'm not too sure if initiating a connection in the command runner is a good pattern or an anti-pattern.

cmd/ocm-backplane/accessrequest/createAccessRequest.go
wanghaoran1988: @Nikokolas3270 Not good at naming, but how about `approve_access_request`?

pkg/utils/util.go
Nikokolas3270: Agreed, put in the config

cmd/ocm-backplane/accessrequest/createAccessRequest.go
Nikokolas3270: As said this description is (normally) be changed a few milliseconds later. So whatever we put won't matter much. Lets put something short & neutral as it will appear in the JIRA ticket history

cmd/ocm-backplane/accessrequest/getAccessRequest.go
Nikokolas3270: Indeed, the `DefaultOCMInterfaceImpl` could manage the `*ocmsdk.Connection` itself.
The only problem would be: when to `Close()` the connection.
Answer probably being: when the `ocm-backplane` CLI exits.
Remark that not properly closing the connection is probably not that of a big a deal. So ideally we should:
- Add a `Close()` method to `OCMInterface`. In `DefaultOCMInterfaceImpl`, this would close the underlying connection if there is one.
- Add a `defer OCMInterface.Close()` in the main method.
WDYT? Could this go in a separate PR? I already want to cleanup a bit the code used to manage the config :)

cmd/ocm-backplane/accessrequest/createAccessRequest.go
Nikokolas3270: I think it is ok a long as we add some `Close()` method

pkg/utils/jira.go
```diff
@@ -0,0 +1,125 @@
+package utils
```
samanthajayasinghe, line 1: @Nikokolas3270 Thanks for moving JIRA client as an interface. 
Shall we move into `github.com/openshift/backplane-cli/pkg/jira` rather than keeping it in package util?

cmd/ocm-backplane/accessrequest/getAccessRequest.go
samanthajayasinghe: Thanks @Nikokolas3270 for the suggestion, let's add `defer OCMInterface.Close()` into the main method but will do it in separate PR

pkg/utils/jira.go
samanthajayasinghe: As We discussed offline, we will move Jira package in a separate PR

PR #469, opened 2024-07-01
[OSD-19783] Add backplane subcommand to do health check of VPN and proxy

xiaoyu74:
### What type of PR is this?

_(feature)_

Add backplane subcommand to do health check of VPN and proxy connectivity
[More discussions in slack](https://redhat-internal.slack.com/archives/C0326L38PEH/p1719196787530089)

https://issues.redhat.com/browse/OSD-19783

**1. Latest updates resolved all the review comments - 16 July**

-  Refactoring to move the main healthcheck logic to the `pkg` folder
-  Updated `Makefile` to use `mockgen` to generate mock files with the mock interfaces for the test cases
-  Removed the default values defined in the `config.go` to avoid expose internal URL, allowing the end-user to customize the endpoint URLs in the local backplane configuration  
   -  `VPNCheckEndpoint`
   -  `ProxyCheckEndpoint`
-  Added debug logs
-  Split the main `connectivity_checks.go` function into `check_vpn.go` and `check_proxy.go` based on the core functionality.
-  The end-user needs to set the vpn and proxy `check-endpoint` in the local bp configuration, example below:
```
cat ~/.config/backplane/config.json 
{
    "proxy-url": ["http://proxy1.example.com:3128", "http://proxy2.example.com:3128"],
    "vpn-check-endpoint": "http://your-vpn-endpoint.example.com",
    "proxy-check-endpoint": "http://your-proxy-endpoint.example.com"
}
```

**2. Command output examples**

-  Command test with VPN and proxy working properly
```
./ocm-backplane health-check
Checking VPN connectivity...
VPN connectivity check passed!

Checking proxy connectivity...
Getting the working proxy URL ['http://proxy1.example.com:3128'] from local backplane configuration.
Testing connectivity to the pre-defined test endpoint ['https://your-proxy-endpoint.example.com'] with the proxy.
Proxy connectivity check passed!

Checking backplane API connectivity...
Successfully connected to the backplane API!
Backplane API connectivity check passed!
```

-  We can check `VPN or Proxy` connectivity respectively by specifying the `--vpn` or `--proxy` flag
```
./ocm-backplane healthcheck --vpn
Checking VPN connectivity...
VPN connectivity check passed!

./ocm-backplane healthcheck --proxy
Checking proxy connectivity...
Proxy connectivity check passed!
```

- When VPN is disconnected
```
./ocm-backplane healthcheck --proxy
WARN[0000] No VPN interfaces found: [tun tap ppp wg utun]    
VPN connectivity check failed: No VPN interfaces found: [tun tap ppp wg utun]
Note: Proxy connectivity check requires VPN to be connected. Please ensure VPN is connected and try again.
```


cmd/ocm-backplane/healthCheck/health_check.go
```diff
@@ -0,0 +1,72 @@
...
+		if err != nil {
+			fmt.Println("VPN connectivity check failed:", err)
+			if checkProxy {
+				fmt.Println("Note: Proxy connectivity check requires VPN to be connected. Please ensure VPN is connected and try again.")
+			}
+			os.Exit(1)
+		} else {
+			fmt.Println("VPN connectivity check passed!")
+		}
+		fmt.Println()
```
samanthajayasinghe, line 32: is this printing an empty line? 

just an idea, you may able to use go [text/template](https://pkg.go.dev/text/template) package to nicely arrange the output text

cmd/ocm-backplane/healthCheck/health_check.go
```diff
@@ -0,0 +1,72 @@
...
+var (
+	checkVPN   bool
+	checkProxy bool
+)
+
+// HealthCheckCmd is the command for performing health checks
+var HealthCheckCmd = &cobra.Command{
+	Use:   "health-check",
+	Short: "Check VPN and Proxy connectivity",
+	Run: func(cmd *cobra.Command, args []string) {
```
samanthajayasinghe, line 19: It's good to move this as separate function to keep the consistency as other backplane subcommand

pkg/cli/config/config.go
```diff
@@ -17,11 +17,12 @@ import (
...
 // Please update the validateConfig function if there is any required config key added
 type BackplaneConfiguration struct {
-	URL              string  `json:"url"`
-	ProxyURL         *string `json:"proxy-url"`
-	SessionDirectory string  `json:"session-dir"`
-	AssumeInitialArn string  `json:"assume-initial-arn"`
-	PagerDutyAPIKey  string  `json:"pd-key"`
+	URL              string   `json:"url"`
+	ProxyURL         *string  `json:"proxy-url"`
+	ProxyURLs        []string `json:"proxy-urls"`
```
samanthajayasinghe, line 22: Why do we need two configs?

cmd/ocm-backplane/healthCheck/connectivity_checks.go
```diff
@@ -0,0 +1,198 @@
...
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("unexpected status code: %v", resp.StatusCode)
+	}
+	return nil
+}
+
+// Checking if the VPN is connected by examining the routing table.
+func CheckVPNConnectivity() error {
+	// Execute the 'ip route' command to check for routes via VPN interfaces
+	cmd := exec.Command("ip", "route")
```
samanthajayasinghe, line 54: `ip route` may not work on all operating systems. 

It's better to use go [net package](https://pkg.go.dev/net) for derive local IP if needed

cmd/ocm-backplane/healthCheck/connectivity_checks.go
```diff
@@ -0,0 +1,198 @@
...
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("failed to execute ip route command: %v", err)
+	}
+
+	// List of common VPN interfaces
+	vpnInterfaces := []string{"tun0", "tun1", "tun2", "tun3", "tap0", "tap1", "ppp0", "ppp1", "ppp2", "ppp3", "wg0", "wg1"}
```
samanthajayasinghe, line 61: I feel that you can get all interfaces via net package rather than hard-code the list.

samanthajayasinghe (commented):
Thanks for the PR @xiaoyu74 .. It's good if you can add at least 50% unit tests to cover new sub command

cmd/ocm-backplane/healthCheck/health_check.go
xiaoyu74: Yeah, just printing an empty line for better output messages. Let me check `text/template` package to see how it can achieve that. Thanks :)

cmd/ocm-backplane/healthCheck/connectivity_checks.go
xiaoyu74: We discussed to use `ip route` or `ip link` in [thread](https://redhat-internal.slack.com/archives/C0326L38PEH/p1719197385072169?thread_ts=1719196787.530089&cid=C0326L38PEH), but yeah, using the `net` package might be better for cross-platform purposes. Let me check.

cmd/ocm-backplane/healthCheck/connectivity_checks.go
xiaoyu74: > I feel that you can get all interfaces via net package rather than hard-code the list.

Let me check how's `net` package achieve that.

samanthajayasinghe: Iterating network interfaces may not work when backplane-cli is running on a docker/podman container since the container may not be able to get the hosts' networks, but I'm not 100% sure. 
You may be able to test this on `ocm-container`, where backplane-cli is running on the container.

xiaoyu74:
/hold 
Will need to discuss and investigate for this subcommand to work in the ocm-container.

pkg/cli/config/config.go
xiaoyu74: @samanthajayasinghe  - My intention is to load all the Proxy urls from the local backplane configuration just as below:
```
Checking proxy connectivity...
Loading proxy URLs from local backplane configuration:
[http://squid.corp.redhat.com:3128; http://proxy2.squi-001.prod.rdu2.dc.redhat.com:3128]
```
And the current `ProxyURL *string "json:\"proxy-url,omitempty\"" ` field cannot handle multiple proxy URLs directly. 

Given that removing the `ProxyURL` field can causes issues with existing functions, I think it might be better to keep it temporarily to ensure backward compatibility, we can refactor later on.

cmd/ocm-backplane/healthCheck/connectivity_checks.go
xiaoyu74: > Iterating network interfaces may not work when backplane-cli is running on a docker/podman container since the container may not be able to get the hosts' networks, but I'm not 100% sure. You may be able to test this on `ocm-container`, where backplane-cli is running on the container.

As I tested in a container based on a custom build image (to have my bp-cli with health-check subcommand included) . The `ip route` works to check the VPN from host network, however, as a pre-requisite, we need to install `iproute` binary for the container and also run the container with `--net=host`

Since I've reported the `ocm-container` issue and it has got fixed, I can test with the actual `ocm-container` to see if it's all good. ([reported issue about ocm-container](https://redhat-internal.slack.com/archives/CFJD1NZFT/p1720411712611059))

cmd/ocm-backplane/healthCheck/connectivity_checks.go
xiaoyu74: > `ip route` may not work on all operating systems.
> 
> It's better to use go [net package](https://pkg.go.dev/net) for derive local IP if needed

I think twice, actually, using `ip route` is indeed a common approach for Linux, and a similar command can be used for macOS (netstat -rn or route get default),  assuming our `backplane-cli` mainly focus on Linux and MacOS.

I did some research, looks golang `net` package doesn't provide direct access to routing table information. However, we probably can use `net` to check for network interfaces and their IP addresses, which can be indicative of a VPN connection (if specific IP ranges or interfaces like tun, ppp, or tap are present), it also close for my current logic via checking the VPN interfaces from `ip route` output.

cmd/ocm-backplane/login/login.go
```diff
@@ -296,6 +296,10 @@ func runLogin(cmd *cobra.Command, argv []string) (err error) {
 			return fmt.Errorf("cannot connect to backplane API URL, check if you need to use a proxy/VPN to access backplane: %v", connErr)
 		}
 
+		// Add suggestion to run connectivity health check if login fails
```
bmeng, line 299: Can we use `logger` here instead of `fmt.Print` so that we can adjust the level easily?

pkg/cli/config/config.go
samanthajayasinghe: I guess it's good to keep `proxy-URL` to accept multiple URLs, which keeps the backward compatibility.
If it really need, we will refactor it via another card/PR  
What do you think?

pkg/cli/config/config.go
```diff
@@ -90,11 +90,10 @@ func GetBackplaneConfiguration() (bpConfig BackplaneConfiguration, err error) {
...
 
-	// proxyURL is required
-	proxyInConfigFile := viper.GetStringSlice("proxy-url")
-	proxyURL := bpConfig.getFirstWorkingProxyURL(proxyInConfigFile)
-	if proxyURL != "" {
-		bpConfig.ProxyURL = &proxyURL
+	// Load proxy URLs as a slice of strings
+	bpConfig.ProxyURLs = viper.GetStringSlice("proxy-url")
+	if len(bpConfig.ProxyURLs) > 0 {
+		bpConfig.ProxyURL = &bpConfig.ProxyURLs[0]
```
bmeng, line 96: does this mean we only use the first specified proxy?

cmd/ocm-backplane/healthCheck/connectivity_checks.go
samanthajayasinghe: I'm not a fan of executing exec OS commands via go, but if it's needed, then go with it.
( pls don't block `login` or `cloud` cmd with exec(`ip route`) )

cmd/ocm-backplane/healthCheck/connectivity_checks.go
samanthajayasinghe: I'm not too sure that, it's a UX to ask users to build images as CLI needs 
but you can warn users if `route` is not available in the base OS and fails silently.

cmd/ocm-backplane/healthCheck/connectivity_checks.go
bmeng: maybe use `ip link` instead of `ip route` to detect the vpn?
`ip link show type tun` would be able to show the vpn is connected or not.

bmeng: so we do not need to loop the interfaces below as well

pkg/cli/config/config.go
xiaoyu74: > does this mean we only use the first specified proxy?

I am going to revert this part as below to only keep the `proxyURL` field and updated my `CheckProxyConnectivity()` function accordingly.
```
// proxyURL is required
	proxyInConfigFile := viper.GetStringSlice("proxy-url")
	proxyURL := bpConfig.getFirstWorkingProxyURL(proxyInConfigFile)
	if proxyURL != "" {
		bpConfig.ProxyURL = &proxyURL
	}
```
So, we will only use the **first working proxy** defined in the local backplane config.

cmd/ocm-backplane/healthCheck/connectivity_checks.go
xiaoyu74: > maybe use `ip link` instead of `ip route` to detect the vpn? `ip link show type tun` would be able to show the vpn is connected or not.

@bmeng - Actually, both `ip route` and `ip link` shows the type `tun`, in the output of the `ip route` below confirm that traffic to these networks is routed through the tun0 interface, indicating an active VPN connection.
```
ip route
10.0.0.0/8 via 10.64.136.1 dev tun0 proto static metric 50
10.64.136.0/23 dev tun0 proto kernel scope link src 10.64.136.4 metric 50
```

For `ip link` output:
```
3: enp0s31f6: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000

200: tun0: <POINTOPOINT,MULTICAST,NOARP,UP,LOWER_UP> mtu 1360 qdisc fq_codel state UNKNOWN mode DEFAULT group default qlen 500
    link/none
```
It shows `UNKNOWN` state on `tun0` interface and looks **no obvious benefits** to check VPN connectivity compared to `ip route`, if there is no golang standard library to better achieve that, we may keep to use `ip route`, WDYT?

cmd/ocm-backplane/healthCheck/connectivity_checks.go
xiaoyu74: As discussed with Sam, let's hold to consider the bp-cli health-check in ocm-container at this stage (update the backplane soruce page for clarification) cc: @bmeng

samanthajayasinghe:
Thanks @xiaoyu74 for moving base logic to `/pkg/health` and using mocks add unit tests .. 
Can you pls update readMe file with the new command ?
https://github.com/openshift/backplane-cli/blob/16993591c759db8286194361e6d26df4341606d8/README.md?plain=1#L58

and pls add a new section and explain how this command works

README.md
```diff
@@ -325,6 +326,58 @@ No issue if only stdout is redirected.
...
+
+### Pre-settings
+The end-user needs to set the VPN and Proxy check-endpoints in the local backplane configuration first:
+```
+cat ~/.config/backplane/config.json 
+{
+    "proxy-url": ["http://squid.corp.redhat.com:3128", "http://proxy2.squi-001.prod.rdu2.dc.redhat.com:3128"],
+    "vpn-check-endpoint": "https://gitlab.cee.redhat.com/service/backplane-api",
+    "proxy-check-endpoint": "https://api.internal-redhat.com/healthz"
+}
```
samanthajayasinghe, line 340: Pls don't add these values to README

README.md
```diff
@@ -325,6 +326,58 @@ No issue if only stdout is redirected.
...
+
+### How to use it
+- Running healthcheck by default
+```
+./ocm-backplane healthcheck
+Checking VPN connectivity...
+VPN connectivity check passed!
+
+Checking proxy connectivity...
+Getting the working proxy URL ['http://squid.corp.redhat.com:3128'] from local backplane configuration.
```
samanthajayasinghe, line 355: here too

README.md
```diff
@@ -325,6 +326,58 @@ No issue if only stdout is redirected.
...
+Testing connectivity to the pre-defined test endpoint ['https://api.backplane.openshift.com/healthz'] with the proxy.
```
samanthajayasinghe, line 356: pls remove sensitive URL's

PR #484, opened 2024-07-16
Feat: Get version from build info

feichashao:
### What type of PR is this?

Feature

Some users might want to install `ocm-backplane`(bp-cli) via
```
$ go install github.com/openshift/backplane-cli/cmd/ocm-backplane@latest
```
this will result in `ocm-backplane version` output a blank result, and also a warning message to ask for upgrade when login.

This PR follow a similar logic as https://github.com/golangci/golangci-lint/blob/ee37ef317f56c821d8333505c95d49bc26efbc7f/cmd/golangci-lint/main.go#L30 to get the version info of the binary:
- if the Version is given by Gorelease (the user download the binary), it will print the Version.
- if the Version is blank (the user self build by `go install`), it will print the version info from the buildinfo.

https://issues.redhat.com/browse/OSD-19781

You can have a test using my repo
https://github.com/feichashao/backplane-cli/releases/tag/v0.1.37

```
$ go install github.com/feichashao/backplane-cli/cmd/ocm-backplane@latest
$ /Users/feichashao/go/bin/ocm-backplane version 
0.1.37
```

#### Guidelines
 
#### Test coverage checks  
- [x] Added unit tests

- [x] Ran unit tests locally


pkg/info/info.go
```diff
@@ -55,5 +57,33 @@ var (
...
+	// get the binary version of the current build from available sources
+	GetBuildVersion() string
+}
+
+type DefaultInfoServiceImpl struct {
+}
+
+func (i *DefaultInfoServiceImpl) GetBuildVersion() string {
+	// If the Version is set by Goreleaser, return it directly.
+	if len(Version) > 0 {
```
samanthajayasinghe, line 76: Nice to use `Version != ""` since Version is `string` type, but the same logic and small performance improvement

pkg/info/info_test.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"runtime/debug"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/openshift/backplane-cli/pkg/info"
+)
+
+// Mock implementation of debug.ReadBuildInfo
+func mockReadBuildInfoAvail() (*debug.BuildInfo, bool) {
```
samanthajayasinghe, line 12: It's great that you program to an interface, and you may able to generate the mock class rather than manually define it.
ex : https://github.com/openshift/backplane-cli/blob/16993591c759db8286194361e6d26df4341606d8/Makefile#L118C2-L118C132
```
#step-1 
mockgen -destination=./pkg/info/mocks/infoServiceMock.go -package=mocks github.com/openshift/backplane-cli/pkg/info InfoService

#step-2 
make mock-gen

# step-3 
# return what-ever-value you need 
mockInfoService.EXPECT().GetBuildVersion().Return("v1.2.3").Times(1)
``` 
if you see any issue with above steps , pls ping me and I'm happy to help :)

samanthajayasinghe (commented):
Thanks @feichashao for the PR and nicely implemented with an interface.. added minor nits and happy to discuss

pkg/info/info.go
feichashao: Thanks! Changed to != "" in the new commit.

pkg/info/info_test.go
feichashao: Thank you for the suggestion and detailed steps @samanthajayasinghe !

In the new commit, I added the mock-gen to generate a mock interface.

In the test cases, I would like to test the logic in function `GetVersion()` (renamed from `GetBuildVersion()` in the new commit). 

The `GetVersion()` will call `ReadBuildInfo()` to get the `buildInfo` from the golang build.  In normal case, `ReadBuildInfo` points to a library function `debug.ReadBuildInfo` (runtime/debug) directly. For testing, I point `ReadBuildInfo` to a self-defined function to mock the desired output. I didn't make an interface for `debug.ReadBuildInfo`.

Not sure if that's the best practice. Would like you hear your suggestions.  Thanks!

feichashao:
Hi @samanthajayasinghe , improved testing and rebased. Please help another review, thanks!

PR #486, opened 2024-07-18
inform user to login to OCM if the backplane failed due to that

bmeng:
### What type of PR is this?

_(feature)_

OCM offline token disabled, backplane login would fail if OCM not logged in, inform the user to do the ocm login if it is not when executing the backplane commands

_Resolves #OSD-21603_

#### Guidelines
 
#### Test coverage checks  
- [ ] Added unit tests

- [ ] Ran unit tests locally

### DIFF
--- a/pkg/ocm/ocmWrapper.go
+++ b/pkg/ocm/ocmWrapper.go
@@ -4,9 +4,11 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"os"
 	"strings"
 
-	"github.com/openshift-online/ocm-cli/pkg/ocm"
+	ocmocm "github.com/openshift-online/ocm-cli/pkg/ocm"
+	ocmurls "github.com/openshift-online/ocm-cli/pkg/urls"
 	ocmsdk "github.com/openshift-online/ocm-sdk-go"
 	acctrspv1 "github.com/openshift-online/ocm-sdk-go/accesstransparency/v1"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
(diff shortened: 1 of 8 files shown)


xiaoyu74:
Thanks for the PR

1. The functionality works all good:
```
 ./ocm-backplane login $CLUSTER_ID
ERRO[0000] failed to create OCM connection: please make sure you have logged into OCM, use "ocm login --use-auth-code --url $ENV" to login

./ocm-backplane cloud console $CLUSTER_ID
ERRO[0000] failed to create OCM connection: please make sure you have logged into OCM, use "ocm login --use-auth-code --url $ENV" to login
```

2. I was thinking to check the test case coverage, but looks all the subcommands don't have the test case to cover the previous function  `ocmConnection, err := ocmsdk.NewConnection().Build()` to initialize OCM connection, so we don't have to update the test case for the new function `SetupOCMConnection()` based on that with the additional notification. 

**LGTM.**

pkg/ocm/ocmWrapper.go
```diff
@@ -45,9 +46,26 @@ type DefaultOCMInterfaceImpl struct{}
...
 
+func (o *DefaultOCMInterfaceImpl) SetupOCMConnection() (*ocmsdk.Connection, error) {
+	ocmNotLoggedInMessage := "Not logged in"
+
+	// Setup connection at the first try
+	connection, err := ocm.NewConnection().Build()
+	if err != nil {
+		if strings.Contains(err.Error(), ocmNotLoggedInMessage) {
+			return nil, fmt.Errorf("please make sure you have logged into OCM, " +
+				"use \"ocm login --use-auth-code --url $ENV\" to login ")
```
xiaoyu74, line 57: Just a nit:

```suggestion
    return nil, fmt.Errorf("Failed to create OCM connection: please ensure you are logged into OCM using the command \"ocm login --use-auth-code --url $ENV\"")
```

pkg/ocm/ocmWrapper.go
```diff
@@ -45,9 +46,26 @@ type DefaultOCMInterfaceImpl struct{}
 
 var DefaultOCMInterface OCMInterface = &DefaultOCMInterfaceImpl{}
 
+func (o *DefaultOCMInterfaceImpl) SetupOCMConnection() (*ocmsdk.Connection, error) {
+	ocmNotLoggedInMessage := "Not logged in"
+
+	// Setup connection at the first try
+	connection, err := ocm.NewConnection().Build()
```
samanthajayasinghe, line 53: pls hook `OCM_URL` env hack for fixing the reported issue
```
urlenv := os.Getenv("OCM_URL")
	if urlenv != "" {
		url, _ := urls.ResolveGatewayURL(urlenv, nil)
		os.Setenv("OCM_URL", url)
	}
```

samanthajayasinghe, line 53: Can we cache the connection in `DefaultOCMInterfaceImpl` struct for optimized performance?

bmeng: yeah, I was about to include this part into the code... not finished yet

bmeng: As currently, in each function, it will close the connection in the end. So the connection itself it is reusable.
Will create a new card to handle this cached connection case.

pkg/ocm/ocmWrapper.go
bmeng: The "Failed to create OCM connection:" will be included in the OCM side error message, I just update the second part

pkg/ocm/ocmWrapper.go
```diff
@@ -35,19 +37,51 @@ type OCMInterface interface {
...
+		os.Setenv("OCM_URL", ocmURL)
+		logger.Debugf("reset the OCM_URL to %s", ocmURL)
+	}
+
+	// Setup connection at the first try
+	connection, err := ocmocm.NewConnection().Build()
+	if err != nil {
+		if strings.Contains(err.Error(), ocmNotLoggedInMessage) {
+			return nil, fmt.Errorf("please ensure you are logged into OCM by using the command " +
+				"\"ocm login --use-auth-code --url $ENV\"")
```
samanthajayasinghe, line 73: Is it always true? 
According to ocm login docs 
```
ocm login --help
.....
      --use-auth-code          Login using OAuth Authorization Code. This should be used for most cases where a browser is available. See --use-device-code for remote hosts and containers.
      --use-device-code        Login using OAuth Device Code. This should only be used for remote hosts and containers where browsers are not available. See --use-auth-code for all other sce
```

bmeng: I am not 100% percent sure what would be the use case for --use-device-code.
But I think we can just simply mention to use `ocm login`

pkg/ocm/ocmWrapper.go
xiaoyu74: All good then.

PR #547, opened 2024-10-25
Accept deprecation warnings from the server and warn the user

smarthall:
### What type of PR is this?

feature

This will allow us to add code in the server to warn the user when their client is old and may be blocked soon.

This is the client side code for https://issues.redhat.com/browse/OSD-19766

#### Guidelines
 
#### Test coverage checks  
- [x] Added unit tests

- [x] Ran unit tests locally

### DIFF
--- a/cmd/ocm-backplane/login/login.go
+++ b/cmd/ocm-backplane/login/login.go
@@ -24,6 +24,7 @@ import (
 	"github.com/openshift/backplane-cli/pkg/backplaneapi"
 	"github.com/openshift/backplane-cli/pkg/cli/config"
 	"github.com/openshift/backplane-cli/pkg/cli/globalflags"
+	"github.com/openshift/backplane-cli/pkg/info"
 	"github.com/openshift/backplane-cli/pkg/jira"
 	"github.com/openshift/backplane-cli/pkg/login"
 	"github.com/openshift/backplane-cli/pkg/ocm"
@@ -518,6 +519,11 @@ func doLogin(api, clusterID, accessToken string) (string, error) {
 		}
 
 		return "", err
+	}
+
+	err = backplaneapi.CheckResponseDeprecation(resp)
+	if errors.Is(err, backplaneapi.ErrDeprecation) {
+		logger.Warnf("The server indicated that backplane-cli version %s is deprecated. Please update as soon as possible.", info.DefaultInfoService.GetVersion())
 	}
 
 	if resp.StatusCode != http.StatusOK {
--- a/pkg/backplaneapi/clientUtils.go
+++ b/pkg/backplaneapi/clientUtils.go
@@ -37,6 +37,7 @@ func makeClientOptions(accessToken string) BackplaneApi.ClientOption {
 		client.RequestEditors = append(client.RequestEditors, func(ctx context.Context, req *http.Request) error {
 			req.Header.Add("Authorization", "Bearer "+accessToken)
 			req.Header.Set("User-Agent", "backplane-cli"+info.Version)
+			req.Header.Set("Backplane-Version", info.DefaultInfoService.GetVersion())
 			return nil
 		})
 		return nil
--- a/pkg/backplaneapi/deprecation.go
+++ b/pkg/backplaneapi/deprecation.go
@@ -1,0 +1,18 @@
+package backplaneapi
+
+import (
+	"errors"
+	"net/http"
+)
+
+const deprecationMsg = "server indicated that this client is deprecated"
+
+var ErrDeprecation = errors.New(deprecationMsg)
+
+func CheckResponseDeprecation(r *http.Response) error {
+	if r.Header.Get("Deprecated-Client") == "true" {
+		return ErrDeprecation
+	}
+
+	return nil
+}
--- a/pkg/backplaneapi/deprecation_test.go
+++ b/pkg/backplaneapi/deprecation_test.go
@@ -1,0 +1,40 @@
+package backplaneapi_test
+
+import (
+	"net/http"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/openshift/backplane-cli/pkg/backplaneapi"
+)
+
+var _ = Describe("backplaneapi/deprecation", func() {
+	It("Returns ErrDeprecation when the 'Deprecated-Client' header is present.", func() {
+		r := http.Response{Header: http.Header{}}
+
+		r.Header.Add("Deprecated-Client", "true")
+
+		err := backplaneapi.CheckResponseDeprecation(&r)
+
+		Expect(err).To(Equal(backplaneapi.ErrDeprecation))
+	})
+
+	It("Returns nil when the 'Deprecated-Client' header is not present.", func() {
+		r := http.Response{Header: http.Header{}}
+
+		err := backplaneapi.CheckResponseDeprecation(&r)
+
+		Expect(err).To(BeNil())
+	})
+
+	It("Returns nil when the 'Deprecated-Client' header is not 'true'.", func() {
+		r := http.Response{Header: http.Header{}}
+
+		r.Header.Add("Deprecated-Client", "false")
+
+		err := backplaneapi.CheckResponseDeprecation(&r)
+
+		Expect(err).To(BeNil())
+	})
+})
--- a/cmd/ocm-backplane/cloud/console.go
+++ b/cmd/ocm-backplane/cloud/console.go
@@ -125,7 +125,7 @@ func runConsole(cmd *cobra.Command, argv []string) (err error) {
 	if consoleArgs.backplaneURL != "" { // Overwrite if parameter is set
 		backplaneConfiguration.URL = consoleArgs.backplaneURL
 	}
-	logger.Infof("Using backplane URL: %s\n", backplaneConfiguration.URL)
+	logger.Infof("Using backplane URL: %s\n", consoleArgs.backplaneURL)
 
 	// Initialize OCM connection
 	ocmConnection, err := ocm.DefaultOCMInterface.SetupOCMConnection()
--- a/pkg/backplaneapi/backplaneapi_suite_test.go
+++ b/pkg/backplaneapi/backplaneapi_suite_test.go
@@ -1,0 +1,13 @@
+package backplaneapi_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestBackplaneapi(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Backplaneapi Suite")
+}


cmd/ocm-backplane/login/login.go
```diff
@@ -520,6 +521,11 @@ func doLogin(api, clusterID, accessToken string) (string, error) {
 		return "", err
 	}
 
+	err = deprecation.CheckResponseDeprecation(resp)
+	if err != nil {
```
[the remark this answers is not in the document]
smarthall: Good call, I'll make this change now.

feichashao:
Thank you for this PR and thank you for including unit tests!

I see there's a repeat pattern,
```
err = deprecation.CheckResponseDeprecation(resp)
	if err != nil {
		logger.Warnf("The server indicated that this version of backplane-cli is deprecated, please update as soon as possible.")
	}
```
do you think we can re-use it? Maybe a function like `CheckResponseDeprecationAndWarn`?

I am also thinking, shall we just put this in the `login`/`cloud`/`session` command? Feels noisy if the user just want to "ignore".

smarthall:
> I see there's a repeat pattern,
> 
> ```
> err = deprecation.CheckResponseDeprecation(resp)
> 	if err != nil {
> 		logger.Warnf("The server indicated that this version of backplane-cli is deprecated, please update as soon as possible.")
> 	}
> ```
> 
> do you think we can re-use it? Maybe a function like `CheckResponseDeprecationAndWarn`?

I didn't want to make the deprecation package responsible for writing to screen and left that to the command itself. I suppose I could add a deprecation package to the `cmd/backplane-api` namespace. Is that how you see this working?

> I am also thinking, shall we just put this in the `login`/`cloud`/`session` command? Feels noisy if the user just want to "ignore".

I went for everything that touched the backplane-api, but you are right, if we warn this much people will just stop noticing it anyway. I'll reduce it to just those three.

pkg/deprecation/deprecation.go
```diff
@@ -0,0 +1,16 @@
+package deprecation
```
bmeng, line 1: Can we put this under the pkg/util? I feel it is not big enough to be a separated package

pkg/backplaneapi/clientUtils.go
```diff
@@ -32,16 +32,19 @@ var (
...
 )
 
-func (s *DefaultClientUtilsImpl) MakeRawBackplaneAPIClientWithAccessToken(base, accessToken string) (BackplaneApi.ClientInterface, error) {
-	co := func(client *BackplaneApi.Client) error {
+func makeClientOptions(accessToken string) BackplaneApi.ClientOption {
+	return func(client *BackplaneApi.Client) error {
 		client.RequestEditors = append(client.RequestEditors, func(ctx context.Context, req *http.Request) error {
 			req.Header.Add("Authorization", "Bearer "+accessToken)
 			req.Header.Set("User-Agent", "backplane-cli"+info.Version)
+			req.Header.Set("Backplane-Version", info.DefaultInfoService.GetVersion())
```
bmeng, line 40: The version should be included in the above line with `info.Version`?

cmd/ocm-backplane/managedJob/getManagedJob.go
```diff
@@ -91,6 +94,11 @@ func newGetManagedJobCmd() *cobra.Command {
 					return err
 				}
 
+				err = deprecation.CheckResponseDeprecation(resp)
+				if errors.Is(err, deprecation.ErrDeprecation) {
+					logger.Warnf("The server indicated that this version of backplane-cli is deprecated, please update as soon as possible.")
```
samanthajayasinghe, line 99: It seems this API deprecates error message duplicates in several places in this PR. Can the error be handled in a single place? 
May be utilize this func : https://github.com/openshift/backplane-cli/blob/594bdf05cc83373296422dda5275974cf5c82121/pkg/utils/util.go#L137

cmd/ocm-backplane/managedJob/logsManagedJob.go
```diff
@@ -101,6 +104,11 @@ func newLogsManagedJobCmd() *cobra.Command {
 				return err
 			}
 
+			err = deprecation.CheckResponseDeprecation(resp)
+			if errors.Is(err, deprecation.ErrDeprecation) {
+				logger.Warnf("The server indicated that this version of backplane-cli is deprecated, please update as soon as possible.")
```
samanthajayasinghe, line 109: Duplicates error block, and this may be avoided with the previous comment

cmd/ocm-backplane/script/describeScript.go
```diff
@@ -89,6 +92,11 @@ func newDescribeScriptCmd() *cobra.Command {
 				return err
 			}
 
+			err = deprecation.CheckResponseDeprecation(resp)
+			if errors.Is(err, deprecation.ErrDeprecation) {
+				logger.Warnf("The server indicated that this version of backplane-cli is deprecated, please update as soon as possible.")
```
samanthajayasinghe, line 97: same here

pkg/backplaneapi/clientUtils.go
```diff
@@ -32,16 +32,19 @@ var (
 	DefaultClientUtils ClientUtils = &DefaultClientUtilsImpl{}
 )
 
-func (s *DefaultClientUtilsImpl) MakeRawBackplaneAPIClientWithAccessToken(base, accessToken string) (BackplaneApi.ClientInterface, error) {
-	co := func(client *BackplaneApi.Client) error {
+func makeClientOptions(accessToken string) BackplaneApi.ClientOption {
```
samanthajayasinghe, line 35: Is there any reason to implement this function without a constructor pattern?
```
ex: 
func (s *DefaultClientUtilsImpl)makeClientOptions(accessToken string) ... {
 s.ClientOption = ....
}
```

pkg/backplaneapi/clientUtils.go
```diff
@@ -53,39 +56,21 @@ func (s *DefaultClientUtilsImpl) MakeRawBackplaneAPIClientWithAccessToken(base,
...
-		})
-		return nil
-	}
+func makeRawBackplaneAPIClientWithAccessTokenCustomProxy(server string, accessToken string, proxyURL string) (BackplaneApi.ClientInterface, error) {
+	co := makeClientOptions(accessToken)
 
-	if proxyURL != nil {
-		proxy, err := url.Parse(*proxyURL)
+	if proxyURL != "" {
+		proxy, err := url.Parse(proxyURL)
```
samanthajayasinghe, line 68: This is a good catch :+1:

pkg/backplaneapi/clientUtils.go
```diff
@@ -100,14 +85,7 @@ func (s *DefaultClientUtilsImpl) MakeRawBackplaneAPIClient(base string) (Backpla
...
 func (*DefaultClientUtilsImpl) MakeBackplaneAPIClientWithAccessToken(base, accessToken string) (BackplaneApi.ClientWithResponsesInterface, error) {
-	co := func(client *BackplaneApi.Client) error {
-		client.RequestEditors = append(client.RequestEditors, func(ctx context.Context, req *http.Request) error {
-			req.Header.Add("Authorization", "Bearer "+accessToken)
-			req.Header.Set("User-Agent", "backplane-cli"+info.Version)
-			return nil
-		})
-		return nil
-	}
+	co := makeClientOptions(accessToken)
```
samanthajayasinghe, line 88: I guess it's not good to mix constructor method with non-constructor methods; what if call this as `s.makeClientOptions()`

samanthajayasinghe (commented):
Thanks, @smarthall, for the PR. It's a great initiative to deprecate the old version. Added a couple of nits and suggestions.

xiaoyu74:
> I am also thinking, shall we just put this in the `login`/`cloud`/`session` command? Feels noisy if the user just want to "ignore".

+1, focusing the deprecation warnings on those 3 subcommands makes sense since these are the most frequently used by SREs. Limiting the warnings for those specifically should effectively to notify users.

smarthall:
I removed the code cleanup commits on request from @samanthajayasinghe. They are now in #552

cmd/ocm-backplane/login/login.go
```diff
@@ -261,10 +262,10 @@ func runLogin(cmd *cobra.Command, argv []string) (err error) {
 
 	logger.Debugln("Extracting backplane URL")
 	// Get Backplane URL
-	bpURL := globalOpts.BackplaneURL
-	if bpURL == "" {
-		bpURL = bpConfig.URL
+	if globalOpts.BackplaneURL != "" {
+		bpConfig.URL = globalOpts.BackplaneURL
```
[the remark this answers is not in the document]
smarthall: The problem is that the bpConfig.URL is used later on and I assume that users expect if you provide a new backplane URL on the CLi it overrides the one in the config everywhere. The other option of course is a much bigger refactor so that the bpConfig object is created with the correct URL when an override is given on the command line.

pkg/backplaneapi/clientUtils.go
smarthall: So this also supports `(devel)` and `unknown` in the relevant places rather than being blank. When its blank it will get an older version as we are forced to assume that it is an older client.

pkg/deprecation/deprecation.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"errors"
+	"net/http"
+)
+
+var ErrDeprecation = errors.New("server indicated that this client is deprecated")
```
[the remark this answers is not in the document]
smarthall: Fixed

pkg/deprecation/deprecation_test.go
```diff
@@ -0,0 +1,36 @@
+package deprecation
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/magiconair/properties/assert"
```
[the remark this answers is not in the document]
smarthall: Switched to go builtins like several other tests.

pkg/backplaneapi/clientUtils.go
smarthall: We talked and I cant do this as there is already another function in this file that also does not have a receiver which calls this function.

cmd/ocm-backplane/login/login.go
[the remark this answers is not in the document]
smarthall: Reverted this change

pkg/backplaneapi/deprecation.go
```diff
@@ -0,0 +1,18 @@
+package backplaneapi
```
samanthajayasinghe, line 1: This is a good package .. we should move the api-client also to this package ( as a separate task )

pkg/deprecation/deprecation.go
smarthall: This has been moved

samanthajayasinghe:
@smarthall You may have to to rebase the PR since we bumped the go version : [PR](https://github.com/openshift/backplane-cli/pull/554)

smarthall:
> I am also thinking, shall we just put this in the `login`/`cloud`/`session` command? Feels noisy if the user just want to "ignore".

The `cloud` command doesn't have an easy place to do this, so I'm just going to go for `login` and `session` for now to avoid a larger refactor.

cmd/ocm-backplane/managedJob/logsManagedJob.go
smarthall: Since we changed the plan so this only prints on `login` and `session` so this message is now only in one place.

cmd/ocm-backplane/login/login.go
```diff
@@ -520,6 +520,11 @@ func doLogin(api, clusterID, accessToken string) (string, error) {
 		return "", err
 	}
 
+	err = backplaneapi.CheckResponseDeprecation(resp)
+	if errors.Is(err, backplaneapi.ErrDeprecation) {
+		logger.Warnf("The server indicated that this version of backplane-cli is deprecated, please update as soon as possible.")
```
xiaoyu74, line 525: Do you think it's better to include the current version of `bp-cli` to make it more actionable? e.g
```
logger.Warnf("The server indicated that backplane-cli version %s is deprecated. Please update as soon as possible.", currentVersion)
```

pkg/backplaneapi/deprecation_test.go
```diff
@@ -0,0 +1,40 @@
+package backplaneapi
+
+import (
```
xiaoyu74, line 3: How about to refactor with `Ginkgo BDD Testing Framework` to align with other parts of the bp-cli ?
https://github.com/openshift/backplane-cli/blob/37a6ee99dcd2a1007458afe3d860ac222432a492/pkg/accessrequest/accessRequest_test.go#L12

smarthall: ha! There is a bit of a mix throughout the repository, so I picked the one I preferred. :-P

I've done the refactor in 0e0875ce03ec86e1ff1cb7750734b58395aaf900

smarthall:
MAtching API code is here: https://gitlab.cee.redhat.com/service/backplane-api/-/merge_requests/385

xiaoyu74:
> > I am also thinking, shall we just put this in the `login`/`cloud`/`session` command? Feels noisy if the user just want to "ignore".
> 
> The `cloud` command doesn't have an easy place to do this, so I'm just going to go for `login` and `session` for now to avoid a larger refactor.

Has the deprecation warnings also implemented in the `session` subcommand?  I didn't see the code changes for that accordingly?

smarthall:
> Has the deprecation warnings also implemented in the `session` subcommand? I didn't see the code changes for that accordingly?

@xiaoyu74 The [session command calls the login command](https://github.com/smarthall/backplane-cli/blob/0e0875ce03ec86e1ff1cb7750734b58395aaf900/pkg/cli/session/session.go#L408-L409), so the implementation there works for both.

PR #561, opened 2024-11-06
Print cluster info aftger login

diakovnec:
### What type of PR is this?

feature to print cluster info after backplain login into a cluster

### What this PR does / Why we need it?

Helps to get basic info while troubleshooting 

### Which Jira/Github issue(s) does this PR fix?

(https://issues.redhat.com/browse/OSD-25315)

### Special notes for your reviewer

There a couple of things to mention:

1. It covers 49.7 % of code but fails when `PrintAccessProtectionStatus(clusterID)` is called as its required to mock ocmConnection

` Unexpected call to *mocks.MockOCMInterface.SetupOCMConnection([]) at /home/ydiakov/cards/osd-25315/attem_4/backplane-cli/pkg/login/printClusterInfoAfterLogin.go:40 because: there are no expected calls of the method "SetupOCMConnection" for that receiver`

#### Guidelines
- If it's a new sub-command or new function to an existing sub-command, please cover at least 50% of the code
- If it's a bug fix for an existing sub-command, please cover 70% of the code 
 
#### Test coverage checks  
- [ ] Added unit tests
- [ ] Created jira card to add unit test
- [ ] This PR may not need unit tests

### Pre-checks (if applicable)
- [ ] Ran unit tests locally
- [ ] Validated the changes in a cluster
- [ ] Included documentation changes with PR

PR #578, opened 2024-12-05
Print cluster info by setting a flag in backplane config file.

diakovnec:
### What type of PR is this?

This is an addition to https://github.com/openshift/backplane-cli/pull/561

This PR adds an ability to add a flag in the backplane config and the cluster info will be printed automatically every time without a need to set `--cluster-into` each time

Example
`
{
    "proxy-url": ["http://squid.corp.redhat.com:3128/","http://proxy2.squi-001.prod.rdu2.dc.redhat.com:3128/"],
    "display-cluster-info": true
}
`

[_Resolves #_
](https://issues.redhat.com/browse/OSD-25315)

### DIFF
--- a/cmd/ocm-backplane/login/login_test.go
+++ b/cmd/ocm-backplane/login/login_test.go
@@ -241,7 +241,51 @@ var _ = Describe("Login command", func() {
 		BeforeEach(func() {
 			globalOpts.Manager = false
 			globalOpts.Service = false
+			args.clusterInfo = false
+
 		})
+
+		It("when display cluster info is set to true", func() {
+			err := utils.CreateTempKubeConfig(nil)
+			args.defaultNamespace = "default"
+			args.clusterInfo = true
+			Expect(err).To(BeNil())
+			mockOcmInterface.EXPECT().GetOCMEnvironment().Return(ocmEnv, nil).AnyTimes()
+			mockOcmInterface.EXPECT().GetTargetCluster(testClusterID).Return(trueClusterID, testClusterID, nil)
+			mockOcmInterface.EXPECT().IsClusterHibernating(gomock.Eq(trueClusterID)).Return(false, nil).AnyTimes()
+			mockOcmInterface.EXPECT().GetOCMAccessToken().Return(&testToken, nil)
+			mockClientUtil.EXPECT().MakeRawBackplaneAPIClientWithAccessToken(backplaneAPIURI, testToken).Return(mockClient, nil)
+			mockClient.EXPECT().LoginCluster(gomock.Any(), gomock.Eq(trueClusterID)).Return(fakeResp, nil)
+			mockOcmInterface.EXPECT().GetClusterInfoByID(gomock.Any()).Return(mockCluster, nil).Times(2)
+			mockOcmInterface.EXPECT().SetupOCMConnection().Return(nil, nil)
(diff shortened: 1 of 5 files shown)


cmd/ocm-backplane/login/login_test.go
```diff
@@ -242,6 +242,49 @@ var _ = Describe("Login command", func() {
 			globalOpts.Manager = false
 			globalOpts.Service = false
 		})
+		It("should print a message if DisplayClusterInfo is not set to true", func() {
```
bmeng, line 245: should not print?

bmeng: maybe we can change the testing to positive test as well? so it would be should print when is set to true

diakovnec: Sorry, I am not getting what needs to be changed 
`It("should print a message if DisplayClusterInfo is set to true", func() {`

cmd/ocm-backplane/login/login_test.go
```diff
@@ -241,7 +243,73 @@ var _ = Describe("Login command", func() {
...
+			args.clusterInfo = false
+
+			mockOcmInterface.EXPECT().GetOCMEnvironment().Return(ocmEnv, nil).AnyTimes()
+			mockOcmInterface.EXPECT().GetTargetCluster(testClusterID).Return(trueClusterID, testClusterID, nil)
+			mockOcmInterface.EXPECT().IsClusterHibernating(gomock.Eq(trueClusterID)).Return(false, nil).AnyTimes()
+			mockOcmInterface.EXPECT().GetOCMAccessToken().Return(&testToken, nil)
+			mockClientUtil.EXPECT().MakeRawBackplaneAPIClientWithAccessToken(backplaneAPIURI, testToken).Return(mockClient, nil)
+			mockClient.EXPECT().LoginCluster(gomock.Any(), gomock.Eq(trueClusterID)).Return(fakeResp, nil)
+
+			// Redirect standard output and log output to a buffer
```
bmeng, line 283: I think we may not need to check the stdout or stderr here.
I cannot see much value of checking this in a negative test.

diakovnec: while looking into I've thought about diff scenario when `PrintClusterInfo(clusterID)` returns an error

pkg/cli/config/config.go
```diff
@@ -42,6 +42,7 @@ type BackplaneConfiguration struct {
 	JiraConfigForAccessRequests AccessRequestsJiraConfiguration `json:"jira-config-for-access-requests"`
 	VPNCheckEndpoint            string                          `json:"vpn-check-endpoint"`
 	ProxyCheckEndpoint          string                          `json:"proxy-check-endpoint"`
+	DisplayClusterInfo          bool                            `json:"display-cluster-info"`
```
xiaoyu74, line 45: As all the setting flags in the backplane configuration are currently defined as string values, but `DisplayClusterInfo` is defined as a bool, it would be better to emphasize that end-user should use the boolean value true instead of the string "true" in the `README.md`.

**P.S** This might not be necessary, if you'd like to go a step further, you could implement a small utility function to handle both boolean and string values for `display-cluster-info`, basically, if the value is a string (`"true" or "false"`), it is converted and interpreted as a boolean.

PR #580, opened 2024-12-12
OSD-27228: Modify tests so that they pass on macOS, fix issue with leaking file

joshbranham:
### What type of PR is this?

_bug,cleanup_

I wanted to resolve a bug but was unable to even run the tests for this project on main from my macOS system. This does a few things to fix that:

- Update the assertions for what type of containerEngine you get in the `console` command to work across different runtimes
- Resolves an issue where we would leak a kubeconfig file in `pkg/elevate` because we were relying on the default behavior where a kubeconfig is written to the value of `LocationOfOrigin`

There are a lot of other things I noticed while in here that I would like to cleanup to make testing both easier to reason about, and rely less on local state. We should be using a mock filesystem, for example, and not actually modifying any local state (or relying on it).

Additionally, I removed some tests from `console` that were essentially testing the same behavior, ie a bad container engine passed.

_Resolves #_

I ran `ocm-backplane console` with this locally to confirm all still works. I fundamentally only touched tests.

#### Guidelines
 
#### Test coverage checks  
- [ ] Added unit tests
- [x] This PR may not need unit tests

- [x] Ran unit tests locally

### DIFF
--- a/pkg/elevate/elevate_test.go
+++ b/pkg/elevate/elevate_test.go
@@ -39,9 +39,7 @@ var fakeAPIConfig = api.Config{
 		},
 	},
 	AuthInfos: map[string]*api.AuthInfo{
-		"anonymous": {
-			LocationOfOrigin: "England",
-		},
+		"anonymous": {},
 	},
 	Contexts: map[string]*api.Context{
 		"default/test123/anonymous": {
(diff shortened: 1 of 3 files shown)


pkg/elevate/elevate_test.go
```diff
@@ -39,9 +39,7 @@ var fakeAPIConfig = api.Config{
 		},
 	},
 	AuthInfos: map[string]*api.AuthInfo{
-		"anonymous": {
-			LocationOfOrigin: "England",
```
samanthajayasinghe, line 43: This is a good catch, it always create unwanted file :)

joshbranham: Yeah this one took a bit. There's some code in the k8s kubeconfig package that basically writes the file with the value from `LocationOfOrigin` in the current working directory if you don't tell it otherwise. The real fix was (unfortunately) relying on setting the temporary value for `KUBECONFIG` in the tests. In the future I can come back through here and wire in a fake filesystem which will make it much easier to make assertions without the risk of touching local files at all.

PR #583, opened 2024-12-19
[WIP] adding --replace flag to restart webconsole pod

diakovnec:
### What type of PR is this?

Feature

### What this PR does / Why we need it?

Addresses the issue with the web console pod failing to start if it is already running in a different terminal session

### Which Jira/Github issue(s) does this PR fix?

https://issues.redhat.com/browse/OSD-23459

Implemented as suggested in https://redhat-internal.slack.com/archives/C016S65RNG5/p1715785884588929

### Unit Test Coverage

Before writing the unit test, I would like to see if the functionality fits the purpose

PR #584, opened 2024-12-24
adding beforeStartCleanUp logic to cleanup console pods

diakovnec:
What type of PR is this?
Feature

What this PR does / Why we need it?
Addresses the issue with the web console pod failing to start if it is already running in a different terminal session

Which Jira/Github issue(s) does this PR fix?
https://issues.redhat.com/browse/OSD-23459

Implemented as suggested in https://redhat-internal.slack.com/archives/C016S65RNG5/p1715785884588929

Creating this PR with a different approach as the original requested `--replace` flag is not easily implemented and can be reviewed https://github.com/openshift/backplane-cli/pull/583

### DIFF
--- a/cmd/ocm-backplane/console/console.go
+++ b/cmd/ocm-backplane/console/console.go
@@ -249,7 +249,11 @@ func (o *consoleOptions) run(cmd *cobra.Command, argv []string) error {
 	if err != nil {
 		return err
 	}
-
+	//Perform a cleanup before starting a new console
+	err = o.beforeStartCleanUp(ce)
+	if err != nil {
+		return err
+	}
 	done := make(chan bool)
 	errs := make(chan error)
 
@@ -640,6 +644,35 @@ func (o *consoleOptions) printURL() error {
 				logger.Warnf("failed opening a browser: %s", err)
 			}
 		}()
+	}
+	return nil
+}
+
+func (o *consoleOptions) beforeStartCleanUp(ce containerEngineInterface) error {
+	clusterID, err := getClusterID()
+	if err != nil {
+		return fmt.Errorf("error getting cluster ID: %v", err)
+	}
+	containersToCleanUp := []string{
+		fmt.Sprintf("monitoring-plugin-%s", clusterID),
+		fmt.Sprintf("console-%s", clusterID),
+	}
+
+	logger.Infoln("Starting initial cleanup of containers")
+
+	for _, c := range containersToCleanUp {
+		exist, err := ce.containerIsExist(c)
+		if err != nil {
+			return fmt.Errorf("failed to check if container %s exists: %v", c, err)
+		}
+		if exist {
+			err := ce.stopContainer(c)
+			if err != nil {
+				return fmt.Errorf("failed to stop container %s during the cleanup process: %v", c, err)
+			}
+		} else {
+			logger.Infof("Container %s does not exist, no need to clean up", c)
+		}
 	}
 	return nil
 }
--- a/cmd/ocm-backplane/console/console_test.go
+++ b/cmd/ocm-backplane/console/console_test.go
@@ -113,6 +113,36 @@ var _ = Describe("console command", func() {
 		err = utils.CreateTempKubeConfig(&testKubeCfg)
 		Expect(err).To(BeNil())
 	}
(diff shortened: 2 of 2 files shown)


cmd/ocm-backplane/console/console.go
```diff
@@ -644,6 +648,35 @@ func (o *consoleOptions) printURL() error {
...
+		if err != nil {
+			return fmt.Errorf("failed to check if container %s exists: %v", c, err)
+		}
+		if exist {
+			err := ce.stopContainer(c)
+			if err != nil {
+				return fmt.Errorf("failed to stop container %s during the cleanup process: %v", c, err)
+			}
+		} else {
+			logger.Infof("Container %s does not exist", c)
```
feichashao, line 674:
```suggestion
			logger.Infof("Container %s does not exist, no need to clean up", c)
```
to avoid scaring users.

cmd/ocm-backplane/console/console_test.go
```diff
@@ -113,6 +114,37 @@ var _ = Describe("console command", func() {
 		err = utils.CreateTempKubeConfig(&testKubeCfg)
 		Expect(err).To(BeNil())
 	}
+	Context("Should perform existing pods cleanup before starting the console", func() {
+		It("Should not return an error if no pods are found", func() {
+			setupConfig()
+			createPathPodman()
+			mockOcmInterface.EXPECT().GetClusterInfoByID(clusterID).Return(nil, fmt.Errorf("no pods found")).AnyTimes()
```
feichashao, line 121: Why we need `GetClusterInfoByID` to return a "no pods found here"?

feichashao: I think we don't need this line because the fake podman we are referring to should have no pod running.

cmd/ocm-backplane/console/console.go
```diff
@@ -644,6 +648,35 @@ func (o *consoleOptions) printURL() error {
...
+	for _, c := range containersToCleanUp {
+		exist, err := ce.containerIsExist(c)
+		if err != nil {
+			return fmt.Errorf("failed to check if container %s exists: %v", c, err)
+		}
+		if exist {
+			err := ce.stopContainer(c)
+			if err != nil {
+				return fmt.Errorf("failed to stop container %s during the cleanup process: %v", c, err)
+			}
```
feichashao, line 672: I think the part is un-tested, see if you can improve the test to cover this part?

https://app.codecov.io/gh/openshift/backplane-cli/pull/584?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=openshift

diakovnec: I've added a mock for the ContainerEngineInterface to be able to test a second part of the code. 
I tried to do `mockgen` to generate an auto mock but the original has internal functions only and can't be used outside the package. In order to fix it we would need to refactor entire package and make sure it won't impact something else. 

Hence from here I can see two ways to get forward:
1. Revert back to 50% coverage and merge
2. Keep as it is now with manually added mock

`
containerEngineInterface interface {
	pullImage(imageName string) error
	putFileToMount(filename string, content []byte) error
	stopContainer(containerName string) error
	// some container engines have special handlings for different types of containers
	runConsoleContainer(containerName string, port string, consoleArgs []string, envVars []envVar) error
	runMonitorPlugin(containerName string, consoleContainerName string, nginxConf string, pluginArgs []string) error
	containerIsExist(containerName string) (bool, error)

`

cmd/ocm-backplane/console/console.go
diakovnec: fixed as suggested

cmd/ocm-backplane/console/console.go
feichashao: Thank you for exploring this!

Let's go with "Revert back to 50% coverage and merge".  

The logic to throw an error in the code is straightforward, I think we are ok to not cover it, as the manual mocking could become another maintenance burden. The manual mocking also involved another 3rd-party package, it can be another thing to maintain.

feichashao: We can refactor the container engine interface in the future.  https://issues.redhat.com/browse/OSD-24949

PR #598, opened 2025-01-20
Fix remediation commands

RaphaelBut:
### What type of PR is this?

bug/cleanup, these new commands did not work before.
They were being tested and developed in parallel to backplane-apis remediation feature, which was subject to change.

The backplane feature is now live, so with these commands its possible to create and delete a "remediation" on a cluster.
For more on remediations see https://gitlab.cee.redhat.com/service/backplane-api#remediations

makes the remediation create and delete command work 

https://issues.redhat.com/browse/OSD-27272

#### Guidelines
 
#### Test coverage checks  
- [X] Added unit tests
- [X] This PR may not need unit tests

- [X] Ran unit tests locally
- [X] Validated the changes in a cluster

### DIFF
--- a/cmd/ocm-backplane/remediation/remediation.go
+++ b/cmd/ocm-backplane/remediation/remediation.go
@@ -103,10 +103,7 @@ func runCreateRemediation(args []string, clusterKey string, urlFlag string) erro
 		return err
 	}
 
-	bpURL := globalOpts.BackplaneURL
-	if bpURL == "" {
-		bpURL = bpConfig.URL
-	}
+	bpURL := bpConfig.URL
 
 	clusterID := bpCluster.ClusterID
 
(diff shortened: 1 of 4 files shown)


cmd/ocm-backplane/remediation/remediation.go
```diff
@@ -103,12 +103,15 @@ func runCreateRemediation(args []string, clusterKey string, urlFlag string) erro
 		return err
 	}
 
-	backplaneHost := bpConfig.URL
+	bpURL := globalOpts.BackplaneURL
```
RaphaelBut, line 106: Also there were some issue with setting the correct config options. Can we maybe start tracking something ( maybe already tracked? ) to improve, how its used. I think any common config, flags, env, should be dealt with before going to a specific command. 

So that we do not have to deal with bpURL, backplaneapi client, etc. in each command?

samanthajayasinghe: just for info @RaphaelBut , We should completely get rid of config/cmd options  backplanURL and we should always get it via OCM API 
https://github.com/openshift/backplane-cli/blob/main/pkg/cli/config/config.go#L254

PR #637, opened 2025-03-17
Adds check for additional backplane users at login

iamkirkbater:
### What type of PR is this?

Feature

This adds a check after a successful login to tell the user about other backplane sessions that are currently on the cluster. This should add a bit of a check-and-balance for when MCS is looking at a cluster where SRE might already be paged for it, or conversely if you're responding to a page and see that there are a few other SRE sessions, it might tip you off to look at Service Logs or PD history to see why other SREs may be logged into this cluster already.

Resolves https://issues.redhat.com/browse/OSD-28832

Still a WIP - need to add additional tests and continue to test in different environments.

#### Guidelines
 
#### Test coverage checks  
- [x] Added unit tests

- [x] Ran unit tests locally

### DIFF
--- a/cmd/ocm-backplane/login/login.go
+++ b/cmd/ocm-backplane/login/login.go
@@ -14,6 +14,7 @@ import (
 	logger "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd/api"
 
@@ -334,13 +335,6 @@ func runLogin(cmd *cobra.Command, argv []string) (err error) {
 	// Query backplane-api for proxy url
 	bpAPIClusterURL, err := doLogin(bpURL, clusterID, *accessToken)
 	if err != nil {
-		// If login failed, we try to find out if the cluster is hibernating
-		isHibernating, _ := ocm.DefaultOCMInterface.IsClusterHibernating(clusterID)
-		if isHibernating {
-			// Hibernating, print an error
-			return fmt.Errorf("cluster %s is hibernating, login failed", clusterKey)
-		}
-
 		// Declare helperMsg
 		helperMsg := "\n\033[1mNOTE: To troubleshoot the connectivity issues, please run `ocm-backplane health-check`\033[0m\n\n"
 
@@ -421,7 +415,42 @@ func runLogin(cmd *cobra.Command, argv []string) (err error) {
 		return err
 	}
 
+	// We return without error from here because the user is still successfully logged in
+	// however, we just cannot check for other logged in users for some reason. Therefore,
+	// we should still return that this command exited successfully, as checking for
+	// the other logged in users should not give an SRE the idea they cannot use the credentials.
+
+	logger.Debugln("Checking for other backplane sessions...")
+	cfg, _ := BuildRestConfig(bpAPIClusterURL, accessToken, proxyURL)
+	clientset, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		logger.Errorln("Unable to build kube client from rest config")
+		return nil
+	}
+
+	sessions, err := login.FindOtherSessions(clientset, cfg)
+	if err != nil {
+		logger.Error("Could not check for other sessions.")
+		return nil
+	}
+
+	login.PrintSessions(os.Stdout, sessions)
+
 	return nil
+}
+
+// BuildRestConfig takes a host, token and optional proxy URL and generates a rest config
+func BuildRestConfig(host string, token *string, proxyURL string) (*rest.Config, error) {
+	cfg := &rest.Config{
+		Host:        host,
+		BearerToken: *token,
+	}
+	if proxyURL != "" {
+		cfg.Proxy = func(*http.Request) (*url.URL, error) {
+			return url.Parse(proxyURL)
+		}
+	}
+	return cfg, nil
 }
 
 // GetRestConfig returns a client-go *rest.Config which can be used to programmatically interact with the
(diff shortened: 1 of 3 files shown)


cmd/ocm-backplane/login/login.go
```diff
@@ -334,13 +334,6 @@ func runLogin(cmd *cobra.Command, argv []string) (err error) {
 	// Query backplane-api for proxy url
 	bpAPIClusterURL, err := doLogin(bpURL, clusterID, *accessToken)
 	if err != nil {
-		// If login failed, we try to find out if the cluster is hibernating
```
iamkirkbater, line 337: There's already a check for this on line 325, this check here is redundant.

cmd/ocm-backplane/login/login.go
```diff
@@ -421,9 +414,27 @@ func runLogin(cmd *cobra.Command, argv []string) (err error) {
...
 
+	logger.Debugln("Checking for other backplane sessions...")
+	cfg, _ := BuildRestConfig(bpAPIClusterURL, accessToken, proxyURL)
+	login.CheckForOtherSessions(cfg)
+
 	return nil
 }
 
+// BuildRestConfig takes a host, token and optional proxy URL and generates a rest config
+func BuildRestConfig(host string, token *string, proxyURL string) (*rest.Config, error) {
```
iamkirkbater, line 425: I abstracted this functionality here since it would otherwise `doLogin` again if I reused the `GetRestConfig` or `GetRestConfigWithConn` functions, so that I could just build the rest config directly (because apparently it's not possible to go from an `api.Config` to the `*rest.Config` :/

iamkirkbater:
Output looks like this:

```> make build && ./ocm-backplane login 2hht2q2e34sc7d30pg93i7h3r9mc9vhe
rm -f ocm-backplane
go build -o ocm-backplane ./cmd/ocm-backplane || exit 1
WARN[0001] Your Backplane CLI is not up to date. Please run the command 'ocm backplane upgrade' to upgrade to the latest version  Current version="(devel)" Latest version=0.1.40
WARN[0004] Env KUBE_PS1_CLUSTER_FUNCTION is not detected. It is recommended to set PS1 to learn which cluster you are operating on, refer https://github.com/openshift/backplane-cli/blob/main/docs/PS1-setup.md
There are no other backplane users logged in.```

File: pkg/login/other_logins_check.go [hunk 3]
```diff
@@ -0,0 +1,151 @@
...
+	}
+	meSlice := strings.Split(me, ":")
+	// meSlice should always be something like
+	// system, serviceaccount, openshift-backplane-role-name, username
+	myNS := meSlice[2]
+	myUser := meSlice[3]
+	logger.Debugf("I am %s:%s", myNS, myUser)
+	sessions := map[string]int{}
+
+	for _, role := range backplane_user_namespaces_to_check {
```
xiaoyu74, line 60: Based on the PR description, it indicates that after a successful login, the user (likely MCS?) should be notified if there are existing SRE backplane sessions. This helps MCS determine if an SRE is already paged and investigating an issue on the cluster? 

However, the current check loops through `cee` and `mcs-tier-two` as well. If the goal is to detect only SRE presence, should `mcs-tier-two` and `cee` be included in the check?

pkg/login/other_logins_check.go
```diff
@@ -0,0 +1,151 @@
+package login
```
samanthajayasinghe, line 1: Thanks, @iamkirkbater, for the PR. What if we change the file name to something readable like "user-sesseion-check" . My main thinking is `other-login-checks` is too generic and may not isolate the real use of it, In future, there may be more of these types of checks, and we could modularize it if we have dedicated `check-name`

+1 for adding the unit tests..

pkg/login/other_logins_check.go
```diff
@@ -0,0 +1,151 @@
...
+
+	authenticationapi "k8s.io/api/authentication/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+const (
```
samanthajayasinghe, line 20: Not against hardcoding the team prefix, but this may cause onboarding new teams and supporting the same CLI for fed ramp backplane API's 
If we can fetch the list via the bp-API endpoint, that would be easy to maintain ( Just my thoughts )

pkg/login/other_logins_check.go
iamkirkbater: The goal is for anyone who logs into a cluster with backplane to see any other users who have logged in with backplane who's users have not been cleaned up yet.

So everyone sees everyone, not any other one way around.

pkg/login/other_logins_check.go
iamkirkbater: Is this available as an endpoint on the backplane API already or would that be net new? Happy to use that if it already exists. 

I'll definitely be checking here to make sure this works with FedRAMP clusters here too.

pkg/login/other_logins_check.go
iamkirkbater: Names are 1000% fungible here. I like the suggestions. I got hung up on figuring out a name so I just named it something stupid to move on, I like your suggestions better :)

pkg/login/other_logins_check.go
samanthajayasinghe: There is no current config API yet , but we are planning to add config API (https://issues.redhat.com/browse/OSD-25654 ) 
We can keep it on the bp-cli side for now

iamkirkbater: Makes sense! Once that API endpoint is done then having this be requested dynamically makes total sense!

MitaliBhalla (approved):

PR #682, opened 2025-05-20
Remove filtering the IP whitelist.

bergmannf:
The trusted IPs in OCM are already trust - and should be vetted when introduced.

So for now - as this is breaking production CAD - it seems better to not filter until we can ensure we don't filter out valid users.

### What type of PR is this?

- [X] Bug
- [ ] Feature
- [ ] Documentation
- [ ] Test Coverage
- [ ] Clean Up
- [ ] Others

The current filtering of the already trusted IPs is breaking CAD as it's not using the proxy, but still is a user of the backplane-CLI.

- Related Issue #
- Closes #

#### Guidelines
 
#### Test coverage checks  
- [ ] Added unit tests

- [ ] Ran unit tests locally


cmd/ocm-backplane/cloud/common_test.go
```diff
@@ -269,7 +269,7 @@ var _ = Describe("getIsolatedCredentials", func() {
 			//Only allow 209 IP
 			Expect(policy).To(ContainSubstring("209.10.10.10"))
 			//Not allow 200 IP
```
typeid, line 271: Comment needs to be fixed to be in line with the test

PR #691, opened 2025-05-27
[SREP-644]:  added check logic to assume-initial-arn

diakovnec:
### What type of PR is this?

- [x] Bug
- [x] Test Coverage

This addresses a bug identified and discussed in https://redhat-internal.slack.com/archives/CCX9DB894/p1740008344309409?thread_ts=1739912990.153199&cid=CCX9DB894

bp-cli's initial checking logic [here](https://github.com/openshift/backplane-cli/blob/7212578be1d4fcaff9406944598b7d552d2cfa9f/cmd/ocm-backplane/cloud/common.go#L231-L243), it only sets the assume-initial-arn if it is empty (""), if user manually set an incorrect ARN, it's not empty, then, code skips the override logic and using the incorrect value.

Added logic for checking the correct ARN and returning a clear error message if the wrong ARN has been provided.

example:
```
/home/ydiakov/funcwork/cards/osd-28646/backplane-cli/cmd/ocm-backplane/ocm-backplane cloud console 2eigjr6f5uj8u7l6ie5ql5656ohk1nbr
WARN[0000] Your Backplane CLI is not up to date. Please run the command 'ocm backplane upgrade' to upgrade to the latest version  Current version=0.0.0-20250527071218-9286da22c149+dirty Latest version=0.1.46
WARN[0007] assume-initial-arn in backplane config is not set to a valid payer ARN, using: arn:aws:sts::811685182089:assumed-role/811685182089-sre/ydiakov 
ERRO[0008] failed to get cloud console for cluster 2eigjr6f5uj8u7l6ie5ql5656ohk1nbr: failed to assume role with isolated backplane flow: invalid assume-initial-arn: arn:aws:sts::811685182089:assumed-role/811685182089-sre/ydiakov, must be one of: prod: arn:aws:iam::922711891673:role/SRE-Support-Role, stage: arn:aws:iam::277304166082:role/SRE-Support-Role, int: arn:aws:iam::277304166082:role/SRE-Support-Role 


```

### DIFF
--- a/cmd/ocm-backplane/cloud/common.go
+++ b/cmd/ocm-backplane/cloud/common.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net"
 	"net/url"
+	"slices"
 	"strings"
 
 	//nolint:gosec
@@ -246,7 +247,22 @@ func (cfg *QueryConfig) getIsolatedCredentials(ocmToken string) (aws.Credentials
 			return aws.Credentials{}, errors.New("backplane config is missing required `assume-initial-arn` property")
 		}
 
-		logger.Debugf("set assume-initial-arn to: %s", cfg.BackplaneConfiguration.AssumeInitialArn)
+	} else if !slices.Contains(
+		[]string{
+			productionAssumeInitialArn,
+			stagingAssumeInitialArn,
+			integrationAssumeInitialArn,
+		},
+		cfg.BackplaneConfiguration.AssumeInitialArn,
+	) {
+		logger.Warnf("assume-initial-arn in backplane config is not set to a valid payer ARN, using: %s", cfg.BackplaneConfiguration.AssumeInitialArn)
+		return aws.Credentials{}, fmt.Errorf("invalid assume-initial-arn: %s, must be one of: prod: %s, stage: %s, int: %s",
+			cfg.BackplaneConfiguration.AssumeInitialArn,
+			productionAssumeInitialArn,
+			stagingAssumeInitialArn,
+			integrationAssumeInitialArn,
+		)
+
 	}
 	initialClient, err := StsClient(cfg.BackplaneConfiguration.ProxyURL)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


bmeng:
is there a use case for us to keep this configurable in backplane cli config?

cmd/ocm-backplane/cloud/common.go
```diff
@@ -246,7 +247,19 @@ func (cfg *QueryConfig) getIsolatedCredentials(ocmToken string) (aws.Credentials
 			return aws.Credentials{}, errors.New("backplane config is missing required `assume-initial-arn` property")
 		}
 
-		logger.Debugf("set assume-initial-arn to: %s", cfg.BackplaneConfiguration.AssumeInitialArn)
+	} else if !slices.Contains([]string{
```
samanthajayasinghe, line 250: Minor nit.. The following may improve the readability.  Otherwise, it's hard to find where elseif ends 
```
	} else if !slices.Contains([]string
               {
		productionAssumeInitialArn,
		stagingAssumeInitialArn,
		integrationAssumeInitialArn,
	      }, cfg.BackplaneConfiguration.AssumeInitialArn) 
	{
	 ............
	}
```

samanthajayasinghe:
I have the same feeling as @bmeng , this configuration may be able to be deleted entirely. Users should not be allowed to add random ARN's in backplane configs..

cmd/ocm-backplane/cloud/common.go
diakovnec: I was able to improve readability as below. 

```

 else if !slices.Contains(
		[]string{
			productionAssumeInitialArn,
			stagingAssumeInitialArn,
			integrationAssumeInitialArn,
		},
		cfg.BackplaneConfiguration.AssumeInitialArn,
	) {
		logger.Warnf("assume-initial-arn in backplane config is not set to a valid payer ARN, using: %s", cfg.BackplaneConfiguration.AssumeInitialArn)
		return aws.Credentials{}, fmt.Errorf("invalid assume-initial-arn: %s, must be one of: prod: %s, stage: %s, int: %s",
			cfg.BackplaneConfiguration.AssumeInitialArn,
			productionAssumeInitialArn,
			stagingAssumeInitialArn,
			integrationAssumeInitialArn,
		)

	}

```

I didn't work as suggested due to `!slices.Contains(
		[]string{` its opening brace `{` on the same line

samanthajayasinghe: Thanks for improving the readability @diakovnec , suggested solution is just an example of how elseif mixed with `}` brace

PR #749, opened 2025-08-11
chore: add missing comments for the public functions

bmeng:
add missing comments for the public functions

Generated by Cursor with Claude-4


smarthall:
These all look accurate to me. I think it'd be nice to have the arguments documented too, but this is definitely an improvement.

/lgtm

PR #914, opened 2026-03-30
Draft: Making RO as login default

feichashao:
POC for https://redhat.atlassian.net/browse/SREP-4149


feichashao:
Close this one as we will go with another approach.

PR #974, opened 2026-07-24
fix: simplify config get to read directly from viper

smarthall:
## Summary
- Replaces the `GetBackplaneConfiguration()` call in `config get` with direct viper reads from the config file
- Avoids pulling in the OCM connection logic and proxy-testing side effects that are unnecessary for simply displaying config values
- Handles both string and string-slice formats for `proxy-url`

## Test plan
- [ ] Run `ocm backplane config get url` and verify it prints the configured URL
- [ ] Run `ocm backplane config get proxy-url` with both string and array formats in config
- [ ] Run `ocm backplane config get all` and verify all values display correctly
- [ ] Run `ocm backplane config get` with no config file present (should print empty values)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Configuration values can now be retrieved directly from the saved JSON configuration file (when present).
  * The `get` command output now includes the proxy URL and the GovCloud setting.
  * The `all` option now displays all supported configuration values from the saved JSON configuration.

### DIFF
--- a/cmd/ocm-backplane/config/get.go
+++ b/cmd/ocm-backplane/config/get.go
@@ -29,7 +29,11 @@ func getConfig(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	if _, err = os.Stat(configPath); err == nil {
+	if _, err = os.Stat(configPath); err != nil {
+		if !os.IsNotExist(err) {
+			return fmt.Errorf("unable to access config file %s: %w", configPath, err)
+		}
+	} else {
 		viper.SetConfigFile(configPath)
 		viper.SetConfigType("json")
 		if err := viper.ReadInConfig(); err != nil {
--- a/cmd/ocm-backplane/config/get.go
+++ b/cmd/ocm-backplane/config/get.go
@@ -2,8 +2,11 @@ package config
 
 import (
 	"fmt"
+	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 
 	"github.com/openshift/backplane-cli/pkg/cli/config"
 )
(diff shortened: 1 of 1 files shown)


smarthall:
Okay I've written some Ansible modules for setting up your machine for SREP access, one of the things it does is configure the backplane `proxy-url`. At the moment this command requires ocm to be logged in, this change means we can set that up unattended before the user logs into ocm.

cmd/ocm-backplane/config/get.go
```diff
@@ -21,33 +24,40 @@ func newGetCmd() *cobra.Command {
 }
 
 func getConfig(cmd *cobra.Command, args []string) error {
-	config, err := config.GetBackplaneConfiguration()
+	configPath, err := config.GetConfigFilePath()
```
samanthajayasinghe, line 27: This may overwrite environment configuration settings. For example : [GitHub - backplane-cli config](https://github.com/openshift/backplane-cli/blob/main/pkg/cli/config/config.go#L153).

cmd/ocm-backplane/config/get.go
```diff
@@ -21,33 +24,40 @@ func newGetCmd() *cobra.Command {
...
+		viper.SetConfigType("json")
+		if err := viper.ReadInConfig(); err != nil {
+			return err
+		}
 	}
 
 	switch args[0] {
 	case URLConfigVar:
-		fmt.Printf("%s: %s\n", URLConfigVar, config.URL)
+		fmt.Printf("%s: %s\n", URLConfigVar, viper.GetString(URLConfigVar))
```
xiaoyu74, line 46: Looks like this changes `config get url` from reporting the effective URL to reading only the raw value stored in the config file? 

The exist implementation has the effective URL is resolved from `BACKPLANE_URL` or the current OCM
env, while the file-backed `url` field is already deprecated and may not be present. (e.g with `BACKPLANE_URL` set and no `url` in the config file).

cmd/ocm-backplane/config/get.go
```diff
@@ -21,33 +24,40 @@ func newGetCmd() *cobra.Command {
...
 	case ProxyURLConfigVar:
-		fmt.Printf("%s: %s\n", ProxyURLConfigVar, proxyURL)
+		fmt.Printf("%s: %s\n", ProxyURLConfigVar, strings.Join(viper.GetStringSlice(ProxyURLConfigVar), ", "))
```
xiaoyu74, line 48: Looks like this bypasses the existing `HTTPS_PROXY` binding and the first-working-proxy selection, so the output changes from the effective proxy to the raw list stored in the file?

xiaoyu74:
Thanks Daniel - I agree with the underlying issue that reading a locally available value such as `proxy-url` should not require an authenticated OCM session or trigger unrelated runtime resolution.

My main concern is that look like this implementation fixes that coupling by changing the contract of `config get` for every key: it previously returned effective configuration, but would now return raw file-only values?

Also,  since this changes observable command behavior, could we add focused tests before merging?

smarthall:
I'm going to take a different approach to this, and instead just edit the config file directly.

PR #980, opened 2026-08-11
[ROSAENG-61274] feat: Add a helper to generate yaml to test scripts

feichashao:
### What type of PR is this?

- [ ] fix (Bug Fix)
- [x] feat (New Feature)
- [ ] docs (Documentation)
- [ ] test (Test Coverage)
- [ ] chore (Clean Up / Maintenance Tasks)
- [ ] other (Anything that doesn't fit the above)

### What this PR does / Why we need it?

Assisted by Claude.

This is a helper function to generate yaml files from a managed scripts (with metadata). This is for local test on a staging cluster manually when developing a managed script.

Test steps provided by Claude:

Step 1 — Create mock metadata and script files

metadata.yaml
```
file: script.sh
name: example-test
description: An example test script for validating managed-script behavior
author: testuser
language: bash
rbac:
  clusterRoleRules:
    - verbs: ["get", "list"]
      apiGroups: [""]
      resources: ["pods", "nodes"]
  roles:
    - namespace: openshift-monitoring
      rules:
        - verbs: ["get", "list"]
          apiGroups: [""]
          resources: ["configmaps"]
envs:
  - key: CLUSTER_NAME
    description: Name of the cluster
    optional: false
  - key: DRY_RUN
    description: Enable dry run mode
    optional: true
allowedGroups:
  - SREP
```

script.sh

```
#!/bin/bash
set -euo pipefail

echo "Hello from test script"
echo "Cluster: $CLUSTER_NAME"

echo "Listing pods in openshift-monitoring..."
oc get pods -n openshift-monitoring

echo "Listing nodes..."
oc get nodes
```

Step 2 — Run the render command
```
ocm backplane testjob render \
  -s ./scripts/SREP/example \
  -p CLUSTER_NAME=my-staging-cluster \
  -o test-job.yaml
```

Step 3 — Generated YAML

The command produces a multi-document YAML file containing all the Kubernetes objects needed to run the test script. This matches what the backplane-api's server-side dry-run generates:

```
apiVersion: v1
automountServiceAccountToken: true
kind: ServiceAccount
metadata:
  creationTimestamp: null
  labels:
    managed.openshift.io/backplane-job-canonical-namespace: TEST
    managed.openshift.io/backplane-job-canonical-script-name: example-test
    managed.openshift.io/backplane-job-id: openshift-job-dev-1786430381
    managed.openshift.io/backplane-job-is-test: "true"
  name: openshift-job-dev-1786430381
  namespace: openshift-backplane-managed-scripts
---
[...]

PR #986, opened 2026-08-18
docs:  Update managed-scripts testing method

feichashao:
### What type of PR is this?

- [ ] fix (Bug Fix)
- [ ] feat (New Feature)
- [x] docs (Documentation)
- [ ] test (Test Coverage)
- [ ] chore (Clean Up / Maintenance Tasks)
- [ ] other (Anything that doesn't fit the above)

Update docs for the changes in https://github.com/openshift/backplane-cli/pull/980

Related to ROSAENG-61274

#### Guidelines
 
#### Test coverage checks  
- [ ] Added unit tests

- [ ] Ran unit tests locally
- [ ] Backward compatible

### DIFF
--- a/docs/testing-managed-scripts.md
+++ b/docs/testing-managed-scripts.md
@@ -8,7 +8,7 @@ The recommended way is `ocm backplane testJob render`, which generates the Kuber
 
 
 ## Prerequisite
-You have a non-production ROSA cluster that you own or control, with cluster-admin permissions to create resources in it.
+You have a non-production ROSA cluster that you own or control, with cluster-admin permissions to create resources in it. For example, your own cluster with direct cluster-admin access via IDP, or a staging cluster where you have backplane elevation permissions.
 
 ## Example
 
--- a/docs/testing-managed-scripts.md
+++ b/docs/testing-managed-scripts.md
@@ -6,6 +6,10 @@ Previously this was done with `ocm backplane testJob create/get/logs`, which req
 
 The recommended way is `ocm backplane testJob render`, which generates the Kubernetes YAML (ServiceAccount, RBAC and Pod) for your script **locally** — no backplane API call is made. You then apply it directly on a non-production cluster where you have cluster-admin access.
 
+
+## Prerequisite
+You have a non-production ROSA cluster that you own or control, with cluster-admin permissions to create resources in it.
+
 ## Example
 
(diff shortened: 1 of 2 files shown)


docs/testing-managed-scripts.md
```diff
@@ -0,0 +1,39 @@
...
+
+The recommended way is `ocm backplane testJob render`, which generates the Kubernetes YAML (ServiceAccount, RBAC and Pod) for your script **locally** — no backplane API call is made. You then apply it directly on a non-production cluster where you have cluster-admin access.
+
+## Example
+
+Assuming your draft script lives in a directory that contains a `metadata.yaml` and the script itself (the same layout used in the [managed-scripts](https://github.com/openshift/managed-scripts) repo):
+
+```bash
+# 1. Log in to a non-production cluster with cluster-admin access
+oc login <cluster-api-url>
```
xiaoyu74, line 19: Just to clarify the prerequisite here:  as we don't use backplane to login but `cluster-admin` instead. 

This `cluster-admin` access is cluster-specific, right? So this workflow assumes the script author has a non-prod cluster they own/control and can obtain direct `cluster-admin` credentials rather than picking an random staging fleet cluster as with the old test-script endpoint.

If that's the intended workflow, maybe worth making that explicit in the doc?

feichashao: That's a good point. Yeah, that's intended and it is a safe method for testing scripts.

Let me mention this explicitly in the doc.

feichashao: Add a new commit with a Prerequisite to mention this.

xiaoyu74: one more thought, does this intentionally require every script contributor to have their own non-prod cluster with direct cluster-admin access?

For SRE users who already have Backplane access, `backplane login + elevate` should also work with the rendered YAML, and lets them test on any suitable staging cluster without having to create/own one. 

The render command itself is independent of how the kubeconfig is obtained

feichashao: Yeah, I think as long as the user have cluster-admin access, either backplane elevation or IDP cluster-admin.

Usually MCS or SRE can create their own cluster with cluster-admin, that's the cleanest way for testing. But sometime they can just borrow a cluster with backplane elevation on staging to test, that will be audited.

I pushed a new commit to make the prerequisite more clear about this. PTAL.

